PR #32, opened 2015-06-02
"last" command added

psycho-ir:
By using last command it's possible to get the last stack version number of specific stack_ref.
CI/CD would be easier and less error prone with this command.


hjacobs:
I'm not sure if I want such command, you can already easily script this with "senza list" and output format ("-o") TSV or JSON. Also you are currently only sorting by the version string, i.e. you won't get proper sort order (compare version strings "2" and "10" for example)

sarnowski:
Agree, at least http://semver.org/ should be supported but as it is highly dependend in your project and strategy, there is no correct solution and should be custom as Henning proposed.

psycho-ir:
Yea, correct. I thought version type is int. But if it is not, sorting problem could happen in list too (I will check it tomorrow):
rows.sort(key=lambda x: (x['stack_name'], x['version']))

I agree, As you mentioned this feature is not viable in cli and anybody who is interested can do it in his side.

hjacobs:
closing as discussed

PR #34, opened 2015-06-03
Use EbsOptimized property only with instance types that support it.

alexeyklyukin:
Update the role policy to also be able to set tags on EC2 instances and EBS storage, in order to tag them with postgres roles and cluster names (particularly, for the storage to be able to locate it in case of sudden instance termination).

### DIFF
--- a/senza/templates/postgresapp.py
+++ b/senza/templates/postgresapp.py
@@ -35,7 +35,9 @@ SenzaComponents:
         Maximum: 3
         MetricType: CPU
       InstanceType: {{instance_type}}
+      {{#ebs_optimized}}
       EbsOptimized: True
+      {{/ebs_optimized}}
       BlockDeviceMappings:
         - DeviceName: /dev/xvdk
           Ebs:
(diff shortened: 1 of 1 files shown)


senza/templates/postgresapp.py
```diff
@@ -123,19 +125,39 @@
...
       Policies:
-      - PolicyName: AmazonEC2ReadOnlyAccess
+      - PolicyName: AmazonS3FullAccess
         PolicyDocument:
           Version: "2012-10-17"
           Statement:
           - Effect: Allow
             Action: "s3:*"
             Resource: "*"
+      - PolicyName: AmazonEC2FullAccess
```
hjacobs: really? I don't want to give my PG cluster full EC2 API access...

senza/templates/postgresapp.py
```diff
@@ -123,19 +125,39 @@
           Action: sts:AssumeRole
       Path: /
       Policies:
-      - PolicyName: AmazonEC2ReadOnlyAccess
+      - PolicyName: AmazonS3FullAccess
```
hjacobs: really? I don't want to give my PG cluster full S3 API access...

alexeyklyukin: I think the PolicyName is just a name and real access control is defined by the Effect, Action and Resource. You have a point though, maybe we can restrict the policy to a single bucket and/or EC2/EBS instances. The reason we need S3 access is for WAL-E logs and for the mint bucket, and at least WAL-E bucket should be writable. As for the EC2 and EBS, we only need to assign tags.

PR #36, opened 2015-06-04
Use EBS optimized instances only when they are supported.

alexeyklyukin:


senza/templates/postgresapp.py
```diff
@@ -133,9 +135,22 @@
 '''
 
 
+def ebs_optimized_supported(instance_type):
+    # per http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html
+    return instance_type in ('c1.large', 'c3.xlarge', 'c3.2xlarge', 'c3.4xlarge',
```
hjacobs: is this hard-coded list really necessary? is there no API to use?

alexeyklyukin: There is no up to my knowledge. The closes I could find is the pricing information (http://stackoverflow.com/questions/7334035/get-ec2-pricing-programmatically), but it doesn't contain a list of instances and relies on URLs that are changing soon.

PR #38, opened 2015-06-05
list sorting problem solved

psycho-ir:


senza/cli.py
```diff
@@ -326,7 +326,7 @@ def list_stacks(region, stack_ref, all, output, watch):
                          'creation_time': calendar.timegm(stack.creation_time.timetuple()),
                          'description': stack.template_description})
 
-        rows.sort(key=lambda x: (x['stack_name'], x['version']))
+        rows.sort(key=lambda x: (x['stack_name'], int(x['version'])))
```
hjacobs: this will NOT work, the stack version allows [a-z0-9] characters, i.e. it will throw an exception if version is a string

PR #61, opened 2015-06-24
Added support for other protocols for load balancer health check.

grassator:
Previously it was always `HTTP`.


hjacobs:
Please fix Travis CI (flake8 formatting)

grassator:
Fixed.

Issue #63, opened 2015-06-29
Senza does not recognize referenced Security Groups


drummerwolli:
senza currently only supports Security Groups which are already created beforehand, not the ones created in the Resources section. Should be supported the same way as IAM roles, or is there a reasons why this is not yet implemented?

hjacobs:
:+1: for supporting "Ref" in the Senza definition, please add a unit test for this case, too

hjacobs:
This issue was done and works.

PR #64, opened 2015-06-29
Enable to recognize referenced Security Groups

chutium:
Ref. https://github.com/zalando-stups/senza/issues/63


hjacobs:
Can you add a unit test for this code branch please?

chutium:
@hjacobs unittest done, and it seems both ```get_security_group``` and ```resolve_security_groups``` methods are covered

Issue #71, opened 2015-07-14
Add some special variables that can be used in the templates


lmineiro:
It would be nice to be able to reference "Well Known Variables" to the senza templates like:
- AWS Account ID
- Region
- MainDomain
- TeamID
- ...

We could use them in some of the conventions we've been using like the mint bucket, etcd cluster and many others:

```
      TaupageConfig:
        etcd_discovery_domain: "etcd.{{Variables.MainDomain}}"
        mint_bucket: "zalando-stups-mint-{{Variables.AccountId}}-{{Variables.Region}}"
```

tuxlife:
It was a hidden feature :relaxed:

https://docs.stups.io/en/latest/components/senza.html#senza-info

is it okay for you?

tuxlife:
Live in Version 0.71

PR #75, opened 2015-07-17
Fix for Feature Request #71

tuxlife:
Add Support for follow new Template-Options:
* AccountInfo.Region
* AccountInfo.AccountAlias
* AccountInfo.AccountID
* AccountInfo.TeamID (the part of the AccountAlias after the first -)
* AccountInfo.Domain (the 1st hosted zone)

@hjacobs please review and push to master etc


senza/cli.py
```diff
@@ -250,6 +252,56 @@ def __init__(self, **kwargs):
...
+
+    def __getTeamID(self):
+        attr = getattr(self, '__TeamID', None)
+        if attr is None:
+            team_id = get_account_alias().split('-', maxsplit=1)[-1]
+            setattr(self, '__TeamID', team_id)
+            return team_id
+        return attr
+
+    AccountID = property(fget=__getAccountID)
```
hjacobs: you can use the "property" decorator

senza/cli.py
```diff
@@ -250,6 +252,56 @@ def __init__(self, **kwargs):
...
+        return attr
+
+    def __getRegion(self):
+        return getattr(self, '__Region', None)
+
+    def __getDomain(self):
+        attr = getattr(self, '__Domain', None)
+        if attr is None:
+            conn = boto3.client('route53')
+            domain = conn.list_hosted_zones()['HostedZones'][0]['Name']
```
hjacobs: I guess this will only work for "plain vanilla" STUPS accounts with only one hosted zone..

Issue #83, opened 2015-08-04
Improve senza create command line parameters handling


a1exsh:
Surprisingly, I couldn't find this being brought up before, so here goes.

Senza create doesn't check for extra parameters on the command line.  Nor can it help you if you misplaced a parameter.  And you have to put them in the certain order, dictated by the .yaml file.

Imagine a .yaml where two numeric parameters follow one another: you run

    senza create my.yaml bla-bla-bla 3 60 <image> ...

but you have misplaced the params, and instead of 3 instances with some check timeout of 60 seconds you are now creating 60 instances with 3 seconds timeout!  So to avoid that you have to keep the .yaml file before your eyes in order to check the order of the parameters.

Wouldn't it be more convenient to provide parameters as key=value?  You can then specify them in any order and the command line becomes self-documenting, e.g.:

    senza create my.yaml APPID=bla-bla-bla NODES=3 TIMEOUT=60 IMAGE=<image> ...

I think for backwards compatibility, we could also keep the current positional approach and support a mix of them (like the Python itself does by the way).

Cheers!

hjacobs:
Similar issue: https://github.com/zalando-stups/senza/issues/70

a1exsh:
So I guess #70 could be implemented on top of this.  I'll try to see if I can prepare a patch.

hjacobs:
@a1exsh merged and released (v. 0.67) -- could you do a PR on the documentation, too?

a1exsh:
Cool, thanks - will do.

PR #84, opened 2015-08-04
Handle keyword parameters in create and print cmds

a1exsh:
Positional and keyword parameters can be used together, much like
in Python.  A positional parameter must not follow one specified
with a keyword.

This also checks that no parameters is specified more than once
(either as positional or keyword) and that every parameter is
assigned a value.  Adding default parameter values should be easy
with this in place.


a1exsh:
This is for #83.
Docs patch to follow, if the approach is OK.

hjacobs:
Looks good :-)

PR #87, opened 2015-08-06
Add the ability to provide custom tags to all resources of the Cloud …

feikesteenbergen:
…Formation stack.

To enable discovery of resources in aws we want to use tags to aid in finding our components.
The StackName will differ quite a lot between teams, as it is probably unique within 1 organization.

This patch adds the ability to provide a list of additional tags to be applied to all the resources.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -417,11 +417,16 @@ def create(definition, region, version, parameter, disable_rollback, dry_run, fo
     for name, parameter in data.get("Parameters", {}).items():
         parameters.append([name, getattr(args, name, None)])
 
-    tags = {
+    tags = {}
+    for tag in input["SenzaInfo"].get('Tags', ()):
+        for key, value in tag.items():
+            tags[key] = evaluate_template(value, info, [], args)
+
+    tags.update( {
         "Name": stack_name,
         "StackName": input["SenzaInfo"]["StackName"],
         "StackVersion": version
-    }
+    } )
 
     if "OperatorTopicId" in input["SenzaInfo"]:
         topic = input["SenzaInfo"]["OperatorTopicId"]
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -422,11 +422,11 @@ def create(definition, region, version, parameter, disable_rollback, dry_run, fo
         for key, value in tag.items():
             tags[key] = evaluate_template(value, info, [], args)
 
-    tags.update( {
+    tags.update({
         "Name": stack_name,
         "StackName": input["SenzaInfo"]["StackName"],
         "StackVersion": version
-    } )
+    })
 
     if "OperatorTopicId" in input["SenzaInfo"]:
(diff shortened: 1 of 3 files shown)


senza/cli.py
```diff
@@ -417,11 +417,16 @@ def create(definition, region, version, parameter, disable_rollback, dry_run, fo
     for name, parameter in data.get("Parameters", {}).items():
         parameters.append([name, getattr(args, name, None)])
 
-    tags = {
+    tags = {}
+    for tag in input["SenzaInfo"].get('Tags', ()):
```
hjacobs: This will fail if "Tags" is missing (your are using an empty tuple)
Please add a unit test for this code path.

senza/templates/postgresapp.py
```diff
@@ -24,6 +24,8 @@
     - ImageVersion:
         Description: "Docker image version of spilo."
   {{/docker_image}}
+  Tags:
+    - SpiloCluster: "{{=<% %>=}}{{Arguments.version}}<%={{ }}=%>"
```
hjacobs: tags have no order, so you should use a simple dict (instead of a list of dicts)

tests/test_cli.py
```diff
@@ -491,7 +491,7 @@ def test_create(monkeypatch):
     data = {'SenzaComponents': [{'Config': {'Type': 'Senza::Configuration'}}],
             'SenzaInfo': {'OperatorTopicId': 'my-topic',
                           'Parameters': [{'MyParam': {'Type': 'String'}}, {'ExtraParam': {'Type': 'String'}}, {'DefParam': {'Type': 'String', 'DefaultValue': 'DefValue'}}],
-                          'StackName': 'test'}}
+                          'StackName': 'test', 'Tags': [{'CustomTag': 'CustomValue'}]}}
```
hjacobs: tags have no order, so you should use a simple dict (instead of a list of dicts)

senza/cli.py
```diff
@@ -417,11 +417,17 @@ def create(definition, region, version, parameter, disable_rollback, dry_run, fo
     for name, parameter in data.get("Parameters", {}).items():
         parameters.append([name, getattr(args, name, None)])
 
-    tags = {
+    tags = {}
+    for tag in input["SenzaInfo"].get('Tags', []):
```
hjacobs: I would make "Tags" a dict (tags have no sorting), i.e.

    for key, val in input["SenzaInfo"].get('Tags', {}).items():
        ..

feikesteenbergen: Only lists or strings are allowed in this section of the template, as they are turned into Mappings in the CF template.

    "Template format error: Every Mappings attribute must be a String or a List."

Perhaps it would be wise to allow custom tags, but in a different section of the template?

hjacobs:
I guess your solution is fine and it's at least consistent with the Parameters property

PR #94, opened 2015-08-17
Fixed default parameter values

lmineiro:
CloudFormation templates were being created with `DefaultValue` instead of `Default` resulting in

    {"Error":{"Code":"ValidationError","Message":"Invalid template parameter property 'DefaultValue'","Type":"Sender"},"RequestId":"..."}

I don't think anyone was using the DefaultValue feature or they would be getting the same CF errors. For this reason the quickest and easiest fix is just to rename the Senza syntax to use `Default`.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -267,7 +267,7 @@ def parse_args(input, region, version, parameter):
         for key, config in param.items():
             # collect all allowed keys and default values regardless
             paras[key] = None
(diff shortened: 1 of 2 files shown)


a1exsh:
I was using it, but in a locally-patched senza version, which didn't have the code to pass all the params to CF apparently.

Issue #99, opened 2015-08-21
Allow direct pull of template definitions from HTTP(s)


lmineiro:
Most appliances don't require downloading or installing anything locally apart from the STUPS tooling, namely mai and senza. Usually there are also Docker images deployed to some public registry.

It would make the experience a lot more straightforward to just fetch the template from a remote HTTP(S) endpoint skipping the need to download or git clone first.

e.g.
``senza create https://raw.githubusercontent.com/zalando/stups-cassandra/master/stups-cassandra.yaml green etcd.myteam.domain.com``

lmineiro:
I had a quick look at this during the weekend and it seems that urllib.request.urlopen would do it, although it required some tweaking with local file paths.

tuxlife:
Live in Version 0.73

PR #101, opened 2015-08-24
Added support to load templates from http(s) endpoints

lmineiro:
This should add https://github.com/zalando-stups/senza/issues/99

Python3 urllib doesn't handle local files gracefully and the solution was to prefix file:// in the absence of '://'


lmineiro:
Not sure if I should add a test with remote endpoint (could even be the exact same test template but over HTTPS). Let me know if that's ok

PR #107, opened 2015-08-26
Add support for configuration of load balancer scheme in Senza

ehartung:
**As** a developer
**I want** to create ELBs with scheme 'internal'
**In order to** have ELBs which are only accessible from inside the VPN via their CNAME


hjacobs:
This is not really enough, the subnets also need to be changed

PR #114, opened 2015-09-15
Fixes #113

### LINKED ISSUE #113: Wrong default ELB when no Scheme is defined
When no scheme is defined for an ELB, senza defaults to creating an internet-facing ELB in the private subnet, which is broken and unusable. 

When no scheme is supplied in the yaml, senza should default to set the created ELB scheme in the CloudFormation Template to internal.

(also some documentation about the changed default would be nice)

mo-gr:
Fix wrong ELB config when no scheme is defined in yam. Fixes #113


Jan-M:
Looks ok, although an explicit use of loadbalancer_scheme in the template might be clearer but brings back the danger of unintended overwrite.

Issue #46, opened 2015-06-09
Allow CloudFormation Ref function in TaupageConfig


marco-hoyer:
I would like to create cloudformation resources like SQS queues or RDS databases in my app.yml and handover resource properties directly into the instances configuration. As environment variables f.e. This would need senza to parse the TaupageConfig section and look for "Ref: foo" statements to process prior to base64 encoding.

hjacobs:
:+1:

Jan-M:
DB_HOST: {"Fn::GetAtt" : [DBInstance, Endpoint.Address]}
DB_PORT: {"Fn::GetAtt" : [DBInstance, Endpoint.Port]}

Similar issue with the above config, probably time to implement this feature for "easy" service discovery.

chris-zen:
My use case:
```yaml
Resources:
  ExhibitorBucket:
    Type: AWS::S3::Bucket
SenzaComponents:
  - ExhibitorService:
      Type: Senza::TaupageAutoScalingGroup
      TaupageConfig:
        environment:
          S3_BUCKET: [{Ref: "ExhibitorBucket"}]
          S3_PREFIX: "exhibitor"
```

Translates into:
```
"UserData": {
    "Fn::Base64": "#taupage-ami-config\nenvironment:\n  S3_BUCKET:\n  - Ref: ExhibitorBucket\n  S3_PREFIX: exhibitor\n..."
}
```

I see that it is a complex problem in the sense that the result of the ref is not known until the cloudformation template has been deployed, but the EC2 instance data has to be known in advance, so it is like the dog that bites his cue.

I would like to study the problem and see whether I can contribute a solution or not. Any advice on where to start by ?

chris-zen:
The good news is that CloudFormation is able to handle these cases, so we just need to figure out how to allow it from senza itself.

```json
{
  "MyInstance": {
    "Type": "AWS::EC2::Instance",
    "Properties": {
      "UserData": {
        "Fn::Base64": {
          "Fn::Join": ["",
            [
              "Database=", {"Fn::GetAtt": ["SampleDB", "Endpoint.Address"]},
              ":", {"Ref": "DatabasePort"}
            ]
          ]
        }
      }
    }
  }
}
```

elgalu:
Can we close this? it seems to be done at #123 already

chris-zen:
It was merged some time ago and it has been working fine for me.

PR #123, opened 2015-10-06
Allow AWS functions and Ref in TaupageConfig

chris-zen:
This PR solves the issue #46

It allows to put any valid Fn::* in UserData and Ref in the TaupageConfig property of a Senza::TaupageAutoScalingGroup.

Something like that:
```yaml
SenzaComponents:
  - Exhibitor:
      Type: Senza::TaupageAutoScalingGroup
      TaupageConfig:
        environment:
          S3_BUCKET: {Ref: "ExhibitorBucket"}
          S3_PREFIX: "exhibitor"
Resources:
  ExhibitorBucket:
      Type: AWS::S3::Bucket
```
will expand into:
```json
{"UserData": {"Fn::Base64": {"Fn::Join": ["", [
  "#taupage-ami-config\nenvironment:\n  S3_BUCKET: ",
  {"Ref": "ExhibitorBucket"},
  "\n  S3_PREFIX: exhibitor\n"]]}}}
```
It first transforms the configuration tree so the nodes representing *Ref* or *Fn::* are transformed into a template string such as *{{ json }}*, where *json* is the JSON representation of the AWS function.
Then it is dumped into a string by the *yaml.dump* function. Once we have an string with the special template entries in it, it is iterated to extract fixed strings and AWS functions and compose them in a *Fn::Join*. The *Fn::Join* is **only used** if required, otherwise it just generated an string as it was done before.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -5,6 +5,7 @@ import collections
 from unittest.mock import MagicMock
 import yaml
 import json
+import io
 from senza.cli import cli, handle_exceptions, AccountArguments
 import botocore.exceptions
 from senza.traffic import PERCENT_RESOLUTION, StackVersion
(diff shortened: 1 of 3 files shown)


tuxlife:
@chris-zen can you please create a additional test in `test_cli.py` and a example under example.

like test_print_replace_mustache or test_print_auto

thank you

hjacobs:
@chris-zen thanks for the PR and implementing a long awaited feature :smile:

tests/test_cli.py
```diff
@@ -358,6 +359,103 @@ def my_client(rtype, *args):
...
+        with open('myapp.yaml', 'w') as fd:
+            yaml.dump(data, fd, default_flow_style=False)
+
+        result = runner.invoke(cli, ['print', 'myapp.yaml', '--region=myregion', '123', 'master-mind'],
+                               catch_exceptions=False)
+
+        output = io.StringIO(result.output)
+        output.readline()
+        awsjson = json.load(output)
+
```
tuxlife: Can you please use
```python
    stdout, cfjson = result.output.split('\n', 1)
    assert 'Generating Cloud Formation template.. OK' == stdout
    data = json.loads(cfjson)
```

chris-zen: Hi @tuxlife, you merged before my last commit to address this: 
https://github.com/chris-zen/senza/commit/49ab814a46f4c5666a897ac4107aaec454b13edc

PR #124, opened 2015-10-07
Fix for the PR allowing to use AWS functions in Taupage configuration

chris-zen:
The problem was that I was assuming all the dictionary keys to be an string but in cases like the following it failed. The new code doesn't assume this anymore and also fixes cases where the dictionaries doesn't have keys at all. I have adapted the tests to contemplate the case originating this bug.

```yaml
ports:
  80: 80
```


tests/test_cli.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import os
+import io
```
tuxlife: Why do you use a extra module?

tests/test_cli.py
```diff
@@ -393,13 +395,15 @@ def my_resource(rtype, *args):
         result = runner.invoke(cli, ['print', 'myapp.yaml', '--region=myregion', '123', 'master-mind'],
                                catch_exceptions=False)
 
-    stdout, cfjson = result.output.split('\n', 1)
-    assert 'Generating Cloud Formation template.. OK' == stdout
-    awsjson = json.loads(cfjson)
+    output = io.StringIO(result.output)
```
tuxlife: Why not split?

tests/test_cli.py
chris-zen: io module is required for StringIO

tests/test_cli.py
chris-zen: Just a personal preference so I don't need to duplicate the output contents in memory. Given that performance is not critical here both options would be ok.

tuxlife: You are also duplicate the contents in memory...

a extra module isn't necessary for this task

chris-zen: Hi @tuxlife, *StringIO* just reads the contents of *result.output* as a stream and is used by *json.loads* to parse as if it was done from a file, while *.split* splits *result.output* into two new strings, the first line with the console output plus the duplicated json output. Furthermore the module *io* is a standard Python 3.x module and its use doesn't create new dependencies nor affects performance at all.

Anyway if you prefer to keep consistency across all tests I will revert StringIO usage into split with no problem. Hopefully we'll have this fix merged as soon as possible ;-)

PR #138, opened 2015-11-11
Allow setting loadbalancer_scheme as a command line argument

jmcs:


hjacobs:
Could you check the (really) minor Coverage decrease?

Issue #137, opened 2015-11-11
RateLimitExceeded not handled automatically


marek-obuchowicz:
We experienced problems during deployment, seems that throttling was not handled correctly here:

```
Traceback (most recent call last):
  File "/usr/local/bin/senza", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/senza/cli.py", line 1207, in main
    handle_exceptions(cli)()
  File "/usr/local/lib/python3.4/dist-packages/senza/cli.py", line 245, in wrapper
    func()
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/senza/cli.py", line 944, in status
    'Values': [stack.StackId]}]))
  File "/usr/local/lib/python3.4/dist-packages/boto3/resources/collection.py", line 82, in __iter__
    for page in self.pages():
  File "/usr/local/lib/python3.4/dist-packages/boto3/resources/collection.py", line 165, in pages
    for page in pages:
  File "/usr/local/lib/python3.4/dist-packages/botocore/paginate.py", line 85, in __iter__
    response = self._make_request(current_kwargs)
  File "/usr/local/lib/python3.4/dist-packages/botocore/paginate.py", line 157, in _make_request
    return self._method(**current_kwargs)
  File "/usr/local/lib/python3.4/dist-packages/botocore/client.py", line 310, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.4/dist-packages/botocore/client.py", line 395, in _make_api_call
    raise ClientError(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (RequestLimitExceeded) when calling the DescribeInstances operation: Request limit exceeded.
Service mylounge stack status: 

[...]

PR #139, opened 2015-11-11
Fixes RateLimitExceeded not handled automatically

marek-obuchowicz:
Fixes https://github.com/zalando-stups/senza/issues/137 in functions called by `senza status`

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -3,6 +3,7 @@ import datetime
 import functools
 import time
 import boto3
+
 from botocore.exceptions import ClientError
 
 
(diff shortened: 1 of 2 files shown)


marek-obuchowicz:
Note, it doesn't fix all possible calls, which could result in this proble - it fixes call used in `senza status` which is called (in a loop) by our deployments (and breaks it, pretty dangerous)

hjacobs:
Hmm, I'm not really happy with that approach --- it adds a lot of code and only solves half of the problem (just for the stuff _you_ need).

marek-obuchowicz:
Agreed, it was a quick fix only for our needs - we moved this functionality from senza into our deployment tooling. Issue should remain open, as the timeout is still a possible problem with current senza.

PR #159, opened 2015-12-29
Feature/postgres encrypt values

feikesteenbergen:
To reduce the number of plaintext default passwords in configurations, the Spilo template now allows one to specify a (random) passwords. Optionally they can be stored encrypted if a suitable key is found in KMS.

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -3,6 +3,7 @@ import datetime
 import functools
 import time
 import boto3
+import base64
 from botocore.exceptions import ClientError
 
 
@@ -24,9 +25,13 @@ def get_security_group(region: str, sg_name: str):
             raise
 
 
-def encrypt(region: str, KeyId: str, Plaintext: str):
+def encrypt(region: str, KeyId: str, Plaintext: str, encode=False):
     kms = boto3.client('kms', region)
-    return kms.encrypt(KeyId=KeyId, Plaintext=Plaintext)['CiphertextBlob']
+    encrypted = kms.encrypt(KeyId=KeyId, Plaintext=Plaintext)['CiphertextBlob']
+    if encode:
+        return base64.b64encode(encrypted).decode('utf-8')
+
+    return encrypted
 
 
 def list_kms_keys(region: str, describe=False):
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -28,7 +28,7 @@ def get_security_group(region: str, sg_name: str):
 def encrypt(region: str, KeyId: str, Plaintext: str, b64encode=False):
     kms = boto3.client('kms', region)
     encrypted = kms.encrypt(KeyId=KeyId, Plaintext=Plaintext)['CiphertextBlob']
-    if encode:
+    if b64encode:
         return base64.b64encode(encrypted).decode('utf-8')
 
     return encrypted
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -24,6 +24,20 @@ def get_security_group(region: str, sg_name: str):
             raise
 
 
+def encrypt(region: str, KeyId: str, Plaintext: str):
+    kms = boto3.client('kms', region)
+    return kms.encrypt(KeyId=KeyId, Plaintext=Plaintext)['CiphertextBlob']
+
+
+def list_kms_keys(region: str, describe=False):
+    kms = boto3.client('kms', region)
+    keys = list(kms.list_keys()['Keys'])
+    if describe:
+        for key in keys:
+            key.update(kms.describe_key(KeyId=key['KeyId'])['KeyMetadata'])
+    return keys
+
+
 def resolve_security_groups(security_groups: list, region: str):
     result = []
     for security_group in security_groups:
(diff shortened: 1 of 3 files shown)


senza/templates/postgresapp.py
```diff
@@ -245,6 +263,29 @@ def gather_user_variables(variables, region, account_info):
...
+    prompt(variables, 'pgpassword_superuser', "Password for PostgreSQL superuser [random]", show_default=False,
+           default=generate_random_password, hide_input=True, confirmation_prompt=True)
+    prompt(variables, 'pgpassword_standby', "Password for PostgreSQL user standby [random]", show_default=False,
+           default=generate_random_password, hide_input=True, confirmation_prompt=True)
+    prompt(variables, 'pgpassword_admin', "Password for PostgreSQL user admin", show_default=True,
+           default='admin', hide_input=True, confirmation_prompt=True)
+
+    variables['kms_arn'] = None
+    if click.confirm('Do you wish to encrypt these passwords using kms?', default=False):
+        kms_keys = [k for k in list_kms_keys(region, True) if k['Description'] !=
```
alexeyklyukin: Shouldn't we complain here if there are no KMS keys except for the default one?

feikesteenbergen: Yes, we could do that and raise an exception I suppose

senza/aws.py
```diff
@@ -24,6 +25,24 @@ def get_security_group(region: str, sg_name: str):
             raise
 
 
+def encrypt(region: str, KeyId: str, Plaintext: str, encode=False):
```
a1exsh: `s/encode/b64encode/` ?

senza/aws.py
```diff
@@ -24,6 +25,24 @@ def get_security_group(region: str, sg_name: str):
...
+    kms = boto3.client('kms', region)
+    encrypted = kms.encrypt(KeyId=KeyId, Plaintext=Plaintext)['CiphertextBlob']
+    if encode:
+        return base64.b64encode(encrypted).decode('utf-8')
+
+    return encrypted
+
+
+def list_kms_keys(region: str, describe=False):
```
alexeyklyukin: this function is called only with describe=True, you can remove this parameter altogether (or make True a default)...

senza/aws.py
```diff
@@ -24,6 +25,24 @@ def get_security_group(region: str, sg_name: str):
...
+    kms = boto3.client('kms', region)
+    keys = list(kms.list_keys()['Keys'])
```
a1exsh: Not very likely in practice, but you might want to use pagination here (this will list up to 100 keys and you can go as high as 1000 w/o pagination according to the docs).

senza/templates/postgresapp.py
```diff
@@ -245,6 +263,29 @@ def gather_user_variables(variables, region, account_info):
...
 
+    prompt(variables, 'pgpassword_superuser', "Password for PostgreSQL superuser [random]", show_default=False,
+           default=generate_random_password, hide_input=True, confirmation_prompt=True)
+    prompt(variables, 'pgpassword_standby', "Password for PostgreSQL user standby [random]", show_default=False,
+           default=generate_random_password, hide_input=True, confirmation_prompt=True)
+    prompt(variables, 'pgpassword_admin', "Password for PostgreSQL user admin", show_default=True,
+           default='admin', hide_input=True, confirmation_prompt=True)
+
+    variables['kms_arn'] = None
+    if click.confirm('Do you wish to encrypt these passwords using kms?', default=False):
```
a1exsh: `s/kms/KMS/` ?

senza/templates/postgresapp.py
```diff
@@ -245,6 +263,29 @@ def gather_user_variables(variables, region, account_info):
...
+        kms_keys = [k for k in list_kms_keys(region, True) if k['Description'] !=
+                    'Default master key that protects my EBS volumes when no other key is defined']
```
a1exsh: How reliable is this check?  Can it be language-specific, for example?  Why do we need to filter this out?

senza/templates/postgresapp.py
```diff
@@ -218,6 +235,7 @@ def gather_user_variables(variables, region, account_info):
            default='{}-{}-spilo-app'.format(get_account_alias(), region))
     prompt(variables, 'instance_type', 'EC2 instance type', default='t2.micro')
     variables['hosted_zone'] = account_info.Domain or 'example.com'
+    variables['add_replica_loadbalancer'] = click.confirm('Do you want a replica ELB?', default=False)
```
a1exsh: Does this change belong to this commit?

senza/templates/postgresapp.py
feikesteenbergen: Not very, I was hoping to be able to use the key alias (more reliable), but it's not part of the description of the key given by boto3. Will investigate further.

hjacobs:
@feikesteenbergen could you add some unit tests for your new functions? I would like to avoid merging any PRs reducing the code coverage :smirk:

hjacobs:
:+1:

PR #163, opened 2016-01-05
added possibility to expose HTTP port on ELB. 

AntonZarutsky:
Usefull in case of applications with internal visibility for vpc. 
default behavior (no any additional parameters) will remain the same - only HTTPS listener

Usage in yaml file:

- AppLoadBalancer:
    Listeners:
      - HTTPS
      - HTTP

### DIFF
--- a/senza/components/elastic_load_balancer.py
+++ b/senza/components/elastic_load_balancer.py
@@ -5,7 +5,7 @@ from clickclick import fatal_error
 from senza.aws import find_ssl_certificate_arn, resolve_security_groups
 
 SENZA_PROPERTIES = frozenset(['Domains', 'HealthCheckPath', 'HealthCheckPort', 'HealthCheckProtocol',
-                              'HTTPPort', 'Name', 'SecurityGroups', 'SSLCertificateId', 'Type'])
+                              'HTTPPort', 'Name', 'SecurityGroups', 'SSLCertificateId', 'Type', 'Listeners'])
 
 
 def get_load_balancer_name(stack_name: str, stack_version: str):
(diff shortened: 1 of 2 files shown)


hjacobs:
Your changes would prevent people from configuring custom listeners, you could already easily define a HTTP listener by overwritting `Listeners` in your Senza definition with custom Cloud Formation objects.

senza/components/elastic_load_balancer.py
```diff
@@ -5,7 +5,7 @@
 from senza.aws import find_ssl_certificate_arn, resolve_security_groups
 
 SENZA_PROPERTIES = frozenset(['Domains', 'HealthCheckPath', 'HealthCheckPort', 'HealthCheckProtocol',
-                              'HTTPPort', 'Name', 'SecurityGroups', 'SSLCertificateId', 'Type'])
+                              'HTTPPort', 'Name', 'SecurityGroups', 'SSLCertificateId', 'Type', 'Listeners'])
```
hjacobs: this will prevent people from overwriting `Listeners` with custom CF code

hjacobs:
See my notes, we cannot merge this atm as it will break compatibility.

senza/components/elastic_load_balancer.py
AntonZarutsky: Thx.
Will check

hjacobs:
I added a note in the documentation on how to overwrite the ELB `Listeners`: http://docs.stups.io/en/latest/components/senza.html#senza-weighteddnselasticloadbalancer

PR #164, opened 2016-01-06
Feature/spilo fixes

feikesteenbergen:


hjacobs:
@feikesteenbergen could you tackle the decreasing code coverage please?

senza/templates/postgresapp.py
```diff
@@ -190,10 +203,10 @@
           Version: "2012-10-17"
           Statement:
           - Effect: Allow
-            Action: s3:*
+            Action:
+              - s3:*
             Resource:
-              - arn:aws:s3:::{{wal_s3_bucket}}/spilo/*
-              - arn:aws:s3:::{{wal_s3_bucket}}
+              - "arn:aws:s3:::zalando-acid-eu-west-1-spilo-app/spilo/{{Arguments.version}}/*"
```
alexeyklyukin: Should be {{wal_s3_bucket}}

senza/templates/postgresapp.py
```diff
@@ -208,6 +221,81 @@
...
+  {{#add_replica_loadbalancer}}
+  SpiloReplicaSG:
+    Type: "AWS::EC2::SecurityGroup"
+    Properties:
+      GroupDescription: "Security Group for the replica ELB"
+      SecurityGroupIngress:
+        - IpProtocol: tcp
+          FromPort: {{postgres_port}}
+          ToPort: {{postgres_port}}
+          CidrIp: 0.0.0.0/0
```
alexeyklyukin: Should we do the same for the replica ELB as we do for the master one? I think securing the database vs unauthorized reads is as important as securing it vs unauthorized writes

senza/templates/postgresapp.py
```diff
@@ -208,6 +221,81 @@
...
+          SourceSecurityGroupId:
+            Fn::GetAtt:
+              - SpiloReplicaSG
+              - GroupId
+        {{/add_replica_loadbalancer}}
+
+        - IpProtocol: tcp
+          FromPort: 0
+          ToPort: 65535
+          SourceSecurityGroupName: "Odd (SSH Bastion Host)"
```
alexeyklyukin: Is the name really hard-coded, or is there a function that gets it?

senza/templates/postgresapp.py
```diff
@@ -208,6 +221,81 @@
...
+    Type: "AWS::EC2::SecurityGroup"
+    Properties:
+      GroupDescription: "Security Group for Spilo members"
+      SecurityGroupIngress:
+        - IpProtocol: tcp
+          FromPort: {{postgres_port}}
+          ToPort: {{postgres_port}}
+          SourceSecurityGroupId:
+            Fn::GetAtt:
+              - SpiloMasterSG
```
alexeyklyukin: What if the master group {{elb_access_cidr}} does not include the address range of spilo members, wouldn't it break the replication?

senza/templates/postgresapp.py
feikesteenbergen: It is hardcoded, will convert it into a function during senza init, together with the Security Group for the zmon-agent.

senza/templates/postgresapp.py
feikesteenbergen: That would be no problem, as the SpiloMemberIngressMembers resource defines that all members of the SpiloMemberSG can communicate freely over all the ports

feikesteenbergen:
Closing (temporarily)

alexeyklyukin:
+1

hjacobs:
:+1:

Issue #166, opened 2016-01-26
create fails with ASG MinSize 0


berndherding:
if AutoScaling Minimum is set to 0, create fails due to missing DesiredCapacity.

berndherding:
from my point of view: fixed.

PR #167, opened 2016-01-26
#166 - create fails on ASG MinSize 0

berndherding:
possible fix for #166, pls review.


senza/components/auto_scaling_group.py
```diff
@@ -146,6 +146,9 @@ def component_auto_scaling_group(definition, configuration, args, info, force, a
         definition["Resources"][asg_name]["Properties"]["MaxSize"] = configuration["AutoScaling"]["Maximum"]
         definition["Resources"][asg_name]["Properties"]["MinSize"] = configuration["AutoScaling"]["Minimum"]
 
+        if configuration["AutoScaling"]["Minimum"] == 0:
+            definition["Resources"][asg_name]["Properties"]["DesiredCapacity"] = 1
```
hjacobs: I would actually do:

```python
definition["Resources"][asg_name]["Properties"]["DesiredCapacity"] = max(configuration["AutoScaling"]["Minimum"], 1)
```

That would be more generic and less a "special" case (avoiding an "if" and avoiding code coverage decrease..)

PR #168, opened 2016-01-27
call list_stacks method more then one time when the NextToken is in results

CyberDem0n:


hjacobs:
Yeah, stupid boto3 is pretty low level, so we missed this manual paging when migrating from boto2...

CyberDem0n:
What about merging it?

hjacobs:
@CyberDem0n released: https://github.com/zalando-stups/senza/releases/tag/1.0.36

PR #170, opened 2016-01-29
Bugfix: Make region dependent calls explicit

feikesteenbergen:
Some features of the Postgresapp (Spilo) template implicitly relied upon region being set in the account_info.
It should in fact have relied upon the region explicitly specified.

Fixes https://github.com/zalando/spilo/issues/44


hjacobs:
:+1: and thanks for the code coverage :smile:

PR #172, opened 2016-02-08
Support adding extra tags to new stacks

jmcs:


hjacobs:
Why do we need this? This looks like a special code path just to add tags --- it could also be achieved via the Senza definitional YAML + custom parameters.

jmcs:
I need it for lizzy. I want to tag the stacks without having to change the yaml file provided in the request.

hjacobs:
@jmcs OK, could you add a test for the added code? I'm not sure why the code coverage is not shown here in the PR... Thanks!

jmcs:
@hjacobs done

PR #190, opened 2016-03-08
output yaml templates

drummerwolli:
please don't merge yet, this is more a proposal. i will finish it after discussion.

The output of the senza cmd's didn't suit all of my usecases. some examples:
- i want to order all stacks by creation time
- i want to restrict the description column in size
- i want the private dns name instead of private ip

therefore i started writing a general approach to configure the output via a yaml file. the current yaml file can look like this:

```yaml
outputFormats:
  list:
    sort_by:
      - CreationTime
      - "Stack Name"
      - version
    columns:
      - col: name
        alias: "Stack Name"
      - col: version
      - col: StackStatus
        alias: status
      - col: TemplateDescription
        alias: description
        max-length: 40
      - col: CreationTime
```
Has to be saved where click is finding it with `click.get_app_dir()` (http://click.pocoo.org/6/api/)

The value of `col:` can be anything from the result set of this function: http://boto3.readthedocs.org/en/latest/reference/services/cloudformation.html#CloudFormation.Client.list_stacks

open issues:
- This currently only works (and imho makes sense) with the default output style. Currently it breaks the json output, i have to investigate into this ...
- I plan to also do this for `senza instances`
- Backward compatibility/default values (currently you have to have this yaml file in place)

let's discuss ;-)

### DIFF
--- a/senza/outputformats.py
+++ b/senza/outputformats.py
@@ -39,17 +39,5 @@ def row_data(stack, output_format):
             else:
                 row_col = column['col']
             row_dict[row_col] = getattr(stack, column['col'])
-            # if 'max-length' in column:
-            #     value_length = len(row_dict[row_col])
-            #     row_dict[row_col] = row_dict[row_col][0:column['max-length']]
-            #     if value_length > column['max-length']:
-            #         row_dict[row_col] = row_dict[row_col] + '...'
     return row_dict
 
-
-# rows.append({'stack_name': stack.name,
-#                          'stack_name': stack.name,
-#                          'version': stack.version,
-#                          'status': stack.StackStatus,
-#                          'creation_time': calendar.timegm(stack.CreationTime.timetuple()),
-#                          'description': stack.TemplateDescription})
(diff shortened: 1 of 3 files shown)


hjacobs:
@drummerwolli the Travis CI build still fails, could you check?

senza/outputformats.py
```diff
@@ -0,0 +1,42 @@
...
+            config = yaml.safe_load(fd)
+    except:
+        config = None
+    config = config or {}
+    return config
+
+
+def load_output_format(cmd):
+    existing_config = load_config('senza')
+    return existing_config['outputFormats'][cmd]
```
hjacobs: I don't really like having snake-case here (in YAML), please use something like "output_formats"

drummerwolli:
thx for the feedback, i will continue working on this as soon as i have time ...

hjacobs:
@drummerwolli I will close the PR for now --- just create a new PR if you find time to work on it.

PR #191, opened 2016-03-11
Add {{AccountInfo.MintBucket}}

elgalu:
Disclaimer: idk how to test this but hoping you get the idea ;)

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -36,6 +36,7 @@ from urllib.parse import quote
 from .traffic import change_version_traffic, print_version_traffic, get_records, get_zone
 from .utils import named_value, camel_case_to_underscore, pystache_render, ensure_keys
 from pprint import pformat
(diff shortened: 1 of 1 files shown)


elgalu:
Thanks @hjacobs for merging.

I tested it today and only seems to work as long as the Mint bucket is on the same AWS region as the app deployed, else I get:

```sh
taupage-init: INFO: Got Redirect while trying to read "myapp-staging/user.json" 
from mint S3 bucket "zalando-stups-mint-123123-eu-central-1". 
Retrying with region eu-central-1, endpoint 
zalando-stups-mint-123123-eu-central-1.s3.amazonaws.com! 
(S3 error message: The bucket you are attempting to access must 
be addressed using the specified endpoint. Please send all future requests to this endpoint.)
```

In this case the bucket is located on `eu-west-1` instead of `eu-central-1`.

Idk how to make this feature work cross region / cross AWS account, is it even possible?

PR #192, opened 2016-03-23
Fix small typo

rafaelcaricio:
While using `senza` I got a error message with a small typo. Just fixing it here.


hjacobs:
Thanks for fixing the typo, but I guess the whole error message is "kind of wrong", maybe we should rephrase to:

"Multiple VPCs are only supported if one VPC is the default VPC (IsDefault=true)!"

http://serverfault.com/questions/509136/how-do-i-set-a-vpc-in-aws-to-default-vpc-true
http://docs.aws.amazon.com/cli/latest/reference/ec2/describe-vpcs.html#examples

PR #199, opened 2016-04-28
Added support for Cross-Stack-Output-Refs in TaupageConfig

whiskeysierra:


senza/aws.py
```diff
@@ -24,6 +24,21 @@ def resolve_referenced_resource(ref: dict, region: str):
...
         else:
             return resource_id
+    elif 'Stack' in ref and 'Output' in ref:
+        cf = boto3.client('cloudformation', region)
+        stack = cf.describe_stacks(
+            StackName=ref['Stack'])['Stacks'][0]
+        if stack['StackStatus'] != 'CREATE_COMPLETE':
+            raise ValueError('Stack "{}" is not ready ("{}")'.format(ref['Stack'], stack['StackStatus']))
+
+        if 'Outputs' not in stack:
```
hjacobs: this can be removed and the following "if" simplyfied by using `stack.get('Outputs', [])`

hjacobs:
Could you please add a description of the use case to the docs and include an example?

hjacobs:
@whiskeysierra thanks for the PR! Could we have a section in http://docs.stups.io/en/latest/components/senza.html on using these references? Maybe ask @lukasniemeier-zalando as his feature regarding cross stack references is also undocumented...

whiskeysierra:
> Could you please add a description of the use case to the docs and include an example?

Where are the docs?

hjacobs:
@whiskeysierra https://github.com/zalando-stups/documentation/blob/master/components/senza.rst

But you can also start a new section in the user guide, the "Senza" page is already quite long.

hjacobs:
@rafaelcaricio @jmcs please review (as you are new maintainers :smirk_cat:)

Issue #185, opened 2016-02-22
search based on stack version and not stack name


drummerwolli:
usecase: i have currently multiple stack versions running with the same version, they all a part of "the big stack", e.g.:
- kafka
- the rest api for incoming events
- a consumer
- zookeeper
- the management interface for kafka

to get a faster overview on e.g. all stacks or instances i could do 
```
senza instances | grep <stack_version>
```
but i think its more conventient maybe to have a command in senza built in to look for stacks having the same version. what do you think?

something like this (only a proposal, idk if this is e.g. backward-compatible):
```
senza instances --version=<stack_version>
```

outime:
Hey, sorry for the mess above but I was just testing stuff and forgot these mentions can't be removed... oh well. I like the suggestion and implemented the feature, let's ping @hjacobs to see what he thinks.

outime:
Correction: no need to implement this feature [as it's already available](https://github.com/zalando-stups/senza/pull/206#issuecomment-218294680), so the issue can be closed.

hjacobs:
Uh, actually @drummerwolli wants something else (filter by version ONLY) which is not currently supported, we _MAY_ consider implementing support for this like:

```
senza instances * v123

outime:
@hjacobs you're right, I wasn't very clever yesterday. Would you accept a PR for that?

hjacobs:
@outime yes, we also may discuss the desired behavior first here --- @jmcs and @rafaelcaricio are the new maintainers of Senza, so they might have an opinion, too :smirk:

drummerwolli:
maybe a search based on a regex for stakcname **and** version would be cool as well ...

jmcs:
-1 on regex. I would implement this with [fnmatch](https://docs.python.org/3.5/library/fnmatch.html) as I think this works better on the command line. @rafaelcaricio @hjacobs if you agree with using fnmatch I would implement this.

drummerwolli:
mh, just thought of regex because in fullstop we have regex now to (with -x option) ...

https://github.com/zalando-stups/fullstop-cli/releases/tag/1.0.22

rafaelcaricio:
👍 for `fnmatch`

rafaelcaricio:
Having a second though --- `fnmatch` is non-brainer for things that match files, but regex for all other cases. So I would go with regex in this case.

jmcs:
[...]

PR #206, opened 2016-05-10
#185 Stack version filtering

outime:


hjacobs:
I think your PR makes no sense as you can already filter by stack version (STACK_REF can be stack name and version):

```
senza instances mystack v123

outime:
A good reason to close it :)

PR #208, opened 2016-05-11
Set the default Spilo instance type to t2.medium

alexeyklyukin:
since t2.micro is too small to even load the docker image.


hjacobs:
:+1:

hjacobs:
:+1:

PR #209, opened 2016-05-11
Regex based stack matching

### LINKED ISSUE #185: search based on stack version and not stack name

jmcs:
This mostly fixes #185 but we still need documentation after.

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -1,9 +1,10 @@
+import base64
+import boto3
 import collections
 import datetime
 import functools
+import re
 import time
-import boto3
-import base64
 from botocore.exceptions import ClientError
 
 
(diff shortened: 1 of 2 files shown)


senza/aws.py
```diff
@@ -255,12 +256,22 @@ def matches_any(cf_stack_name: str, stack_refs: list):
...
+    except ValueError:
+        name = cf_stack_name
+        version = None
+
     for ref in stack_refs:
-        if ref.version and cf_stack_name == ref.cf_stack_name():
-            return True
-        elif not ref.version and (cf_stack_name or '').rsplit('-', 1)[0] == ref.name:
-            return True
+        matches_name = re.match(ref.name + '$', name)
```
rafaelcaricio: Why to append `'$'` to the regex?

jmcs: So searching for `my-app` doesn't automatically match `my-app-db` for example

rafaelcaricio: Then it is a problem of the regex. But I see your point.

rafaelcaricio:
👍

rafaelcaricio:
Needs rebase.

hjacobs:
:+1:

PR #213, opened 2016-05-12
Fix minor issues with pluggable templates.

alexeyklyukin:
- avoid crashing on a module without the description
- make sure a template to plug is always a module (i.e. not a function).

### DIFF
--- a/senza/templates/__init__.py
+++ b/senza/templates/__init__.py
@@ -1,9 +1,13 @@
-from types import ModuleType
+from types import ModuleType, FunctionType
 import pkg_resources
 
 
 def get_template_description(name, module: ModuleType):
     return '{}: {}'.format(name, (module.__doc__ or "").strip())
+
+
+def has_functions(module, names):
+    return all(isinstance(getattr(module, function_name, None), FunctionType) for function_name in names)
 
 
 def get_templates() -> dict:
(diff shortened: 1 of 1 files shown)


jmcs:
:+1:

jmcs:
Another, perhaps better, solution would be to check if modules has attributes `gather_user_variables` and `generate_definition`

alexeyklyukin:
good point, let me add some extra checks.

senza/templates/__init__.py
```diff
@@ -19,5 +23,7 @@ def get_templates() -> dict:
             # ignore bad entry points
             continue
         else:
-            template_modules[e.name] = module
+            # make sure the entry point resolves to a module with the essential interface functions
+            if isinstance(module, ModuleType) and has_functions(module, ('gather_user_variables', 'generate_definition')):
+                template_modules[e.name] = module
```
rafaelcaricio: Would be nice to log in case templates do not conform with expectations. It is good for debugging, etc.

jmcs: I think we can do that when we implement proper verbosity levels.

senza/templates/__init__.py
```diff
@@ -24,6 +24,7 @@ def get_templates() -> dict:
             continue
         else:
             # make sure the entry point resolves to a module with the essential interface functions
-            if isinstance(module, ModuleType) and has_functions(module, ('gather_user_variables', 'generate_definition')):
+            if isinstance(module, ModuleType) and\
```
jmcs: Can you use `()` instead of `\`, as recommended by [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)?

jmcs:
:+1:

jmcs:
Thanks for the PR.

alexeyklyukin:
Thank you for the original feature :-) 👍

PR #215, opened 2016-05-19
Replace error message

rafaelcaricio:
Replace error message based on comment in #192. Makes it more clear to users what is going on.


hjacobs:
:+1:

hjacobs:
:+1:

PR #226, opened 2016-05-31
Warn about insecure images

### LINKED ISSUE #223: Warn about insecure images on creation.
PierOne now exposes security relevant information about images. Senza is already checking the availability of an image an can use the same step to also warn about insecure containers.

PierOne exposes two information: max severity of CVEs for which a fix is available and the max severity of CVEs for which no fix is available. In case of the former, senza should print a warning that you want to deploy a container for which a fix is easily available.

rafaelcaricio:
Fixes #223
Depends on https://github.com/zalando-stups/pierone-cli/pull/38

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ clickclick>=1.0
 pystache
 PyYAML
 dnspython3
-stups-pierone>=1.0.20
+stups-pierone>=1.0.26
 boto3>=1.3.0
 botocore>=1.4.10
 pytest>=2.7.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ clickclick>=1.0
 pystache
 PyYAML
 dnspython3
-stups-pierone>=1.0.26
+stups-pierone>=1.0.34
 boto3>=1.3.0
 botocore>=1.4.10
 pytest>=2.7.3
--- a/senza/components/taupage_auto_scaling_group.py
+++ b/senza/components/taupage_auto_scaling_group.py
@@ -20,16 +20,6 @@ def check_docker_image_exists(docker_image: pierone.api.DockerImage):
             exists = pierone.api.image_exists('pierone', docker_image)
             if exists:
                 image_tag = pierone.api.get_image_tag('pierone', docker_image)
-                if image_tag['severity_fix_available'] not in ['NOT_PROCESSED_YET',
-                                                               'COULDNT_FIGURE_OUT',
-                                                               'NO_CVES_FOUND',
-                                                               'TOO_OLD']:
-                    click.secho(textwrap.dedent('''
-                    You are deploying an image that has *{}* severity
-                    security fixes easily available!  Please check this artifact tag in
-                    pierone and see which software versions you should upgrade to apply
-                    those fixes.
-                    '''.format(image_tag['severity_fix_available'])).replace('\n', ' ').strip(), fg='red', bold=True)
         except pierone.api.Unauthorized:
             msg = textwrap.dedent('''
             Unauthorized: Cannot check whether Docker image "{}" exists in Pier One Docker registry.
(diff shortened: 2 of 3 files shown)


senza/components/taupage_auto_scaling_group.py
```diff
@@ -1,33 +1,42 @@
...
 def check_docker_image_exists(docker_image: pierone.api.DockerImage):
     if 'pierone' in docker_image.registry:
         try:
             exists = pierone.api.image_exists('pierone', docker_image)
+            if exists:
+                image_tag = pierone.api.get_image_tag('pierone', docker_image)
+                if image_tag['severity_fix_available'] not in ['NOT_PROCESSED_YET',
+                                                               'COULDNT_FIGURE_OUT',
+                                                               'NO_CVES_FOUND',
+                                                               'TOO_OLD']:
```
rafaelcaricio: @sarnowski Could you review that those stuatuses are the ones we do **not** want to show the `warning`? I am not sure about the `TOO_OLD` and `NOT_PROCESSED_YET` maybe we want to show the warning for those statuses too.

sarnowski: Yes both should warn the user as well. It doesn't prevent the user from doing it anyway.

hjacobs:
@rafaelcaricio please update https://github.com/zalando-stups/pierone-cli/pull/38

senza/components/taupage_auto_scaling_group.py
```diff
@@ -1,33 +1,42 @@
...
 from senza.components.auto_scaling_group import component_auto_scaling_group
 from senza.docker import docker_image_exists
 from senza.utils import ensure_keys
 
-
 _AWS_FN_RE = re.compile(r"('[{]{2} (.*?) [}]{2}')", re.DOTALL)
 
 
 def check_docker_image_exists(docker_image: pierone.api.DockerImage):
     if 'pierone' in docker_image.registry:
```
hjacobs: this actually will not trigger for OS registry (registry.opensource.zalan.do) --- this check is meant to check whether we need authentication

rafaelcaricio:
@hjacobs done.

senza/components/taupage_auto_scaling_group.py
rafaelcaricio: Hm... right. I will try another approach then.

rafaelcaricio: Updated the approach! Maybe now is even better because we will show a warning when user is trying to deploy an image that is not from Pierone.

rafaelcaricio:
@hjacobs @sarnowski please review.

hjacobs:
@rafaelcaricio can we have a test please?

senza/components/taupage_auto_scaling_group.py
```diff
@@ -1,37 +1,51 @@
...
-
     else:
         exists = docker_image_exists(str(docker_image))
+
+    warn_msg = 'You are deploying an image that was not automatically '
+    'checked for vulnerabilities. Images stored in Pierone are automatically checked!'
+
     if not exists:
         raise click.UsageError('Docker image "{}" does not exist'.format(docker_image))
+    elif 'severity_fix_available' in image_tag:
```
hjacobs: `image_tag` might not be defined?

hjacobs:
And what about OS registry now?

senza/components/taupage_auto_scaling_group.py
rafaelcaricio: :scream: You're right! I am changing 4 projects at same time... I need to focus better.

rafaelcaricio:
@hjacobs Now it handles any docker image registry, if it does not find the pierone endpoints it will just assume it is not pierone. :mask:

senza/components/taupage_auto_scaling_group.py
```diff
@@ -1,38 +1,62 @@
...
 _AWS_FN_RE = re.compile(r"('[{]{2} (.*?) [}]{2}')", re.DOTALL)
 
 
 def check_docker_image_exists(docker_image: pierone.api.DockerImage):
+    token = None
     if 'pierone' in docker_image.registry:
-        try:
-            exists = pierone.api.image_exists('pierone', docker_image)
-        except pierone.api.Unauthorized:
+        token = get_existing_token('pierone')
```
hjacobs: this will not work for service tokens

rafaelcaricio: Why? That is the exact code that pierone-cli was calling inside `get_existing_token`.

rafaelcaricio: How can I make it work?

hjacobs: OK, might be, so it would just work for human users --- you can use zign.api.get_token(...) which tries best effort and also works with service tokens

requirements.txt
```diff
@@ -2,7 +2,7 @@ clickclick>=1.0
 pystache
 PyYAML
 dnspython3
-stups-pierone>=1.0.20
+stups-pierone>=1.0.26
```
hjacobs: this should require >=1.0.34

hjacobs:
:+1:

Issue #100, opened 2015-08-21
Read named arguments from a file, SCM or HTTP(S) endpoint


lmineiro:
For appliances it is more common to use someone else's senza template and end up providing your own parameters.

Sometimes it could make more sense to just feed the values for those parameters directly from a file or directly from some Git or HTTP endpoint.

For ex., to use the stups-cassandra appliance, directly from https://github.com/zalando/stups-cassandra, one could write a personal stups-cassandra-_myteam_.yaml with parameter values specific to _myteam_:

    EtcdDomain: etcd.myteam.example.com
    ScalyrKey: myteam-specific-scalyr-api-key
    ApplicationId: myteam-stups-cassandra-according-to-yourturn

Such parameter files could be used with a special notation like:

    senza create definition.yaml version @stups-cassandra-myteam.yaml

or

    senza create --params stups-cassandra-myteam.yaml definition.yaml version

Along with https://github.com/zalando-stups/senza/issues/99 this could even become something like:

    senza create \
        --params https://raw.githubusercontent.com/myself/project/master/stups-cassandra-myteam.yaml  \
        https://raw.githubusercontent.com/zalando/stups-cassandra/master/stups-cassandra.yaml \
        cluster1

hjacobs:
Just a note regarding your last comment (read params from HTTP endpoint): I guess this would bring along new problems as parameters will mostly be private and could/should not be read via HTTP without authentication.

lmineiro:
We're now doing something very similar to this using the strategy i described in #155 

That approach still makes the templates long and ugly, not to mention that it allowed me to find the 1024 chars limits for a stack's description:
```
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateStack operation: Template format error: 'Description' length is greater than 1024.
Makefile:23: recipe for target 'deploy' failed
```

PR #227, opened 2016-06-01
Possibility to read parameters from file

ramirantala:
This pull request would allow reading parameters from file.

Kind of what was suggested in issue #100 

The benefit would be that you could have environment specific arguments in files like..

senza create --parameter-file **ci**.yaml your-senza.yaml 1 1
senza create --parameter-file **prod**.yaml your-senza.yaml 1 1

It should work like before, so if you are missing parameter from file, that is defined senza.yaml, you need to give it from cli

senza create --parameter-file **prod**.yaml your-senza.yaml 1 1 MissingParam=TheParam

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -568,13 +568,14 @@ def create(definition, region, version, parameter, disable_rollback, dry_run, fo
 @click.argument('version', callback=validate_version)
 @click.argument('parameter', nargs=-1)
 @region_option
+@parameter_file_option
 @click.option('--disable-rollback', is_flag=True, help='Disable Cloud Formation rollback on failure')
 @click.option('--dry-run', is_flag=True, help='No-op mode: show what would be created')
 @click.option('-f', '--force', is_flag=True, help='Ignore failing validation checks')
-def update(definition, region, version, parameter, disable_rollback, dry_run, force):
+def update(definition, region, version, parameter, disable_rollback, dry_run, force, parameter_file):
     '''Update an existing Cloud Formation stack from the given Senza definition file'''
     region = get_region(region)
-    data = create_cf_template(definition, region, version, parameter, force)
+    data = create_cf_template(definition, region, version, parameter, force, parameter_file)
     cf = boto3.client('cloudformation', region)
 
     with Action('Updating Cloud Formation stack {}..'.format(data['StackName'])) as act:
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -604,7 +604,6 @@ def print_cfjson(definition, region, version, parameter, output, force, paramete
 
 def create_cf_template(definition, region, version, parameter, force, parameter_file):
     region = get_region(region)
-    print("V2")
     if parameter_file is not None:
         parameter_from_file = read_parameter_file(parameter_file)
         parameter = parameter + parameter_from_file
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -402,12 +402,17 @@ def parse_args(input, region, version, parameter, account_info):
 def read_parameter_file(parameter_file):
     paras = []
     try:
-        with open(parameter_file, 'r') as ymlfile:
-            cfg = yaml.load(ymlfile)
-        for key, val in cfg.items():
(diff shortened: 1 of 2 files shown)


ramirantala:
senza update fails. Will fix that.

senza/cli.py
```diff
@@ -398,6 +399,18 @@ def parse_args(input, region, version, parameter, account_info):
     return args
 
 
+def read_parameter_file(parameter_file):
+    paras = []
+    try:
+        with open(parameter_file, 'r') as ymlfile:
+            cfg = yaml.load(ymlfile)
+        for key, val in cfg.items():
+            paras.append("%s=%s" % (key, val))
```
jmcs: Use `.format`  instead of the old style `%` (`"{}={}".format(key, value)`)

senza/cli.py
```diff
@@ -398,6 +399,18 @@ def parse_args(input, region, version, parameter, account_info):
...
+    except:
```
jmcs: This exception is too broad and can catch unintended Exceptions (like SystemExit and KeyboardInterrupt).

senza/cli.py
```diff
@@ -398,6 +399,18 @@ def parse_args(input, region, version, parameter, account_info):
     return args
 
 
+def read_parameter_file(parameter_file):
```
jmcs: This function should be unit tested

ramirantala:
Added few tests and improved error catching.

senza/cli.py
```diff
@@ -398,6 +399,23 @@ def parse_args(input, region, version, parameter, account_info):
...
+    paras = []
+    try:
+        ymlfile = open(parameter_file, 'r')
+    except OSError:
+        raise click.UsageError('Can\'t read parameter file "{}"'.format(parameter_file))
+
+    try:
+        cfg = yaml.load(ymlfile)
```
hjacobs: this should use `safe_load`

senza/cli.py
```diff
@@ -581,17 +601,21 @@ def update(definition, region, version, parameter, disable_rollback, dry_run, fo
...
     region = get_region(region)
-    data = create_cf_template(definition, region, version, parameter, force)
+    data = create_cf_template(definition, region, version, parameter, force, parameter_file)
     print_json(data['TemplateBody'], output)
 
 
-def create_cf_template(definition, region, version, parameter, force):
+def create_cf_template(definition, region, version, parameter, force, parameter_file):
     region = get_region(region)
+    if parameter_file is not None:
```
hjacobs: I would just do `if parameter_file:` to make it shorter (we can treat empty or other falsy values same as None here)

senza/cli.py
ramirantala: ok, I will change.

hjacobs:
@ramirantala can you create a PR for the documentation also, please?

ramirantala:
Sure I can. I was kind of waiting if this goes through.

kenden:
Nice! This feature will help.

senza/cli.py
```diff
@@ -155,6 +155,7 @@ def convert(self, value, param, ctx):
 
 region_option = click.option('--region', envvar='AWS_DEFAULT_REGION', metavar='AWS_REGION_ID',
                              help='AWS region ID (e.g. eu-west-1)')
+parameter_file_option = click.option('--parameter-file', help='Config file for params', metavar='PATH')
```
rafaelcaricio: Could we user a better name for `metavar` like `metavar=SENZA_PARAMS_FILE`? Just to not confuse with `PATH` env var.

hjacobs: @rafaelcaricio I think `PATH` is good enough, seeing "--parameter-file PATH" is concise :smile:

hjacobs:
:+1:

senza/cli.py
rafaelcaricio: 👍

hjacobs:
Released: https://github.com/zalando-stups/senza/releases/tag/1.0.66

PR #231, opened 2016-06-10
Ask before creating IAM role

### LINKED ISSUE #224: senza init creates IAM role without asking
Example run:
```
senza init service1.yml                                                                                                                                      
Please select the project template
1) bgapp: Background app with single EC2 instance
2) postgresapp: HA Postgres app, which needs an S3 bucket to store WAL files
3) rediscluster: Elasticache cluster running multiple redis nodes, with replication / HA
4) redisnode: Elasticache node running redis, without replication / HA (for caching)
5) webapp: HTTP app with auto scaling, ELB and DNS
Please select (1-5): 5
Application ID [hello-world]: 
Docker image without tag/version (e.g. "pierone.example.org/myteam/myapp") [stups/hello-world]: 
HTTP port [8080]: 
HTTP health check path [/]: 
EC2 instance type [t2.micro]: 
Did you need OAuth-Credentials from Mint? [y/N]: n
Please select the load balancer scheme
1) internal: only accessible from the own VPC
2) internet-facing: accessible from the public internet
Please select (1-2) [1]: 1
Checking security group app-hello-world.. OK
Security group app-hello-world does not exist. Do you want Senza to create it now? [Y/n]: n
Checking security group app-hello-world-lb.. OK
Security group app-hello-world-lb does not exist. Do you want Senza to create it now? [Y/n]: n
Checking IAM role app-hello-world.. OK
Creating IAM role app-hello-world.. OK
Generating Senza definition file service1.yml.. OK
```

The issue is with the following lines:
```
Checking IAM role app-hello-world.. OK
Creating IAM role app-hello-world.. OK
```

I think we should be asked if we wanted the IAM role to be created, like it happens for Security Groups.

jmcs:
Fixes #224

### DIFF
--- a/senza/templates/_helper.py
+++ b/senza/templates/_helper.py
@@ -2,6 +2,7 @@ import json
 import re
 
 import boto3
+import botocore.exceptions
 import click
 import clickclick
 from click import confirm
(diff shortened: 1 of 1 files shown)


hjacobs:
:+1:

PR #232, opened 2016-06-10
Explicit flag to load reference from file

### LINKED ISSUE #228: Exit code for non-existing file changed to 0
It worked in 1.0.52 like this:

```bash
$ senza patch foo.yml 1 --region eu-central-1 --image=latest 
Error: Could not open file foo.yml: [Errno 2] No such file or directory: 'foo.yml'
$ echo $?
1
$ senza -V
Senza 1.0.52
```

This changed somewhere between 1.0.52 and 1.0.63

```bash
$ senza patch foo.yml 1 --region eu-central-1 --image=latest 
$ echo $?                                                    
0
$ senza -V
Senza 1.0.63
```

jmcs:
Fixes #228.

### DIFF
--- a/senza/arguments.py
+++ b/senza/arguments.py
@@ -1,0 +1,76 @@
+import os
+from urllib.error import URLError
+from urllib.parse import quote
+from urllib.request import urlopen
+
+import yaml
+from click import Choice, IntRange, ParamType, option
+
+
+class DefinitionParamType(ParamType):
+    name = 'definition'
+
+    def convert(self, value, param, ctx):
+        if isinstance(value, str):
+            try:
+                url = (value if '://' in value
+                       else 'file://{}'.format(quote(os.path.abspath(value))))
+                response = urlopen(url)
+                data = yaml.safe_load(response.read())
+            except URLError:
+                self.fail('"{}" not found'.format(value), param, ctx)
+        else:
+            data = value
+        for key in ['SenzaInfo']:
+            if 'SenzaInfo' not in data:
+                self.fail('"{}" entry is missing in YAML file "{}"'.format(key, value),
+                          param, ctx)
+        return data
+
+
+class KeyValParamType(ParamType):
+    """
+    >>> KeyValParamType().convert(('a', 'b'), None, None)
+    ('a', 'b')
+    """
+    name = 'key_val'
+
+    def convert(self, value, param, ctx):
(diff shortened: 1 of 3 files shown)


hjacobs:
I don't like this proposal, I would rather keep the current behavior and maybe check for `.yaml` or `.yml` file extension (i.e. if `STACK_REF` ends with `.yaml` it can be considered pointing to a file).

jmcs:
Explicit is better than implicit, having too much "magic" behaviour can back us to a corner when we need to change things.

hjacobs:
The CLI is IMHO about user experience and convenience --- so having some magic is fine. I just want to press `<TAB>` to autocomplete the YAML file name, adding `-d` is unnecessary.

senza/arguments.py
```diff
@@ -0,0 +1,76 @@
...
+            try:
+                url = (value if '://' in value
+                       else 'file://{}'.format(quote(os.path.abspath(value))))
+                response = urlopen(url)
+                data = yaml.safe_load(response.read())
+            except URLError:
+                self.fail('"{}" not found'.format(value), param, ctx)
+        else:
+            data = value
+        for key in ['SenzaInfo']:
```
rafaelcaricio: Why this line?

senza/references.py
```diff
@@ -0,0 +1,77 @@
...
+    last_stack = None
+    while refs:
+        ref = refs.pop()
+        if last_stack is not None and VERSION_RE.match(ref):
+            stack_refs.append(StackReference(last_stack, ref))
+        else:
+            try:
+                with open(ref) as fd:
+                    data = yaml.safe_load(fd)
+                ref = data['SenzaInfo']['StackName']
```
rafaelcaricio: Probably we should catch a `KeyError` here in case it is a invalid senza definition file.

rafaelcaricio:
I have to agree that the definition file as obligatory first argument is fine and a good thing (without the -d). 

The problem is that some sub-commands have the Senza definition file obligatory and others have it as optional and might be a stack name.

IMHO we should check if the parameter does match the stack name regex and if not try to read as a file and not the other way around. So when it fails not matching the file we can say that it cannot be a stack name and is not a file (return error code 1).

jmcs:
I'll reimplement this using @rafaelcaricio 's idea.

PR #235, opened 2016-06-13
Validate application_id and application_version

### LINKED ISSUE #157: Validate application_id and application_version
Senza's TaupageAutoScalingGroup should validate application_id and application_version as we do in Kio (https://github.com/zalando-stups/kio/blob/master/resources/api/kio-api.yaml#L482).

* application_id must match pattern: "^[a-z][a-z0-9-]*[a-z0-9]$"
* application_version must match pattern: "^[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?$"

Background: sometimes users accidentally put the full Docker image (pierone.example.org/../..:123) into application_version

jmcs:
Fixes #157

### DIFF
--- a/senza/components/taupage_auto_scaling_group.py
+++ b/senza/components/taupage_auto_scaling_group.py
@@ -19,18 +19,16 @@ APPLICATION_ID_RE = re.compile(r"^[a-z][a-z0-9-]*[a-z0-9]$")
 APPLICATION_VERSION_RE = re.compile(r"^[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?$")
 
 
-def check_application_id(app_id: str) -> bool:
+def check_application_id(app_id: str):
     if not APPLICATION_ID_RE.match(app_id):
         raise click.UsageError('Application id must satisfy regular '
                                'expression pattern "{}"'.format(APPLICATION_ID_RE.pattern))
-    return True
 
 
-def check_application_version(version: str) -> bool:
+def check_application_version(version: str):
     if not APPLICATION_VERSION_RE.match(version):
         raise click.UsageError('Application version must satisfy regular '
                                'expression pattern "{}"'.format(APPLICATION_VERSION_RE.pattern))
-    return True
 
 
 def check_docker_image_exists(docker_image: pierone.api.DockerImage):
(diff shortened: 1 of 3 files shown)


prayerslayer:
looks good

senza/components/taupage_auto_scaling_group.py
```diff
@@ -14,6 +14,24 @@
 
 _AWS_FN_RE = re.compile(r"('[{]{2} (.*?) [}]{2}')", re.DOTALL)
 
+# from kio OpenAPI yaml
+APPLICATION_ID_RE = re.compile(r"^[a-z][a-z0-9-]*[a-z0-9]$")
+APPLICATION_VERSION_RE = re.compile(r"^[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?$")
```
rafaelcaricio: This RE does have unnecessary stuff also it matches something like ``1..3.....4`` as a valid version.
 I would recommend using this simplified version ``^[A-Za-z0-9]+([._-]?[A-Za-z0-9]+)*$``

jmcs: I'm using the same on as kio.

rafaelcaricio: Should we propagate the mistake or fix it? :poop:
I will send a PR to kio.

hjacobs: Why is the "mistake" a problem? I personally don't care, it's only about valid character ranges and that it does not start/end with a symbol, having ten dots after each other is stupid, but not illegal.

hjacobs:
:+1:

senza/components/taupage_auto_scaling_group.py
```diff
@@ -14,6 +14,24 @@
...
+
+
+def check_application_id(app_id: str) -> bool:
+    if not APPLICATION_ID_RE.match(app_id):
+        raise click.UsageError('Application id must satisfy regular '
+                               'expression pattern "{}"'.format(APPLICATION_ID_RE.pattern))
+    return True
```
hjacobs: do we need the return? we could get rid of the two lines :smirk:

senza/components/taupage_auto_scaling_group.py
rafaelcaricio: If everyone is fine with the "different" version pattern. OK! 👍

Just wanted to express my opinion -- If it is to validate versions, lets validate it properly.

rafaelcaricio:
👍

senza/components/taupage_auto_scaling_group.py
rafaelcaricio: https://github.com/zalando-stups/kio/pull/66

Issue #233, opened 2016-06-11
improve the validation of the region in command senza status/wait/events/resources...


kenden:
calling
`$ senza status --region senza-stack-template.yml stack1`
or
`$ senza wait --region minion-stack-template.yml stack1`
or
`senza events --region minion-stack-template.yml stack1`
...
(notice I forgot the region)
results in:
```
Unknown Error.
Please create an issue with the content of /var/folders/90/pz80sxl93gl11zjsbmylwmxj42vx03/T/senza-traceback-giep3cgd

cat /var/folders/90/pz80sxl93gl11zjsbmylwmxj42vx03/T/senza-traceback-giep3cgd
Traceback (most recent call last):
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/senza/error_handling.py", line 35, in wrapper
    func()
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/senza/cli.py", line 934, in status
    for stack in sorted(get_stacks(stack_refs, region)):
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/senza/aws.py", line 236, in get_stacks
    results = cf.list_stacks(**kwargs)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/botocore/client.py", line 262, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/user/dev/project1/deploy/.venv/lib/python3.5/site-packages/botocore/client.py", line 541, in _make_api_call
    operation_model, request_dict)
[...]

PR #237, opened 2016-06-15
Add validation of '--region' param

kenden:
Fixes https://github.com/zalando-stups/senza/issues/233

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -153,8 +153,22 @@ class KeyValParamType(click.ParamType):
         return key_val
 
 
+def validate_version(ctx, param, value):
+    if not VERSION_PATTERN.match(value):
+        raise click.BadParameter('Version must satisfy regular expression pattern "[a-zA-Z0-9]+"')
+    return value
+
+
+def validate_region(ctx, param, value):
+    """Validate Click region param parameter."""
+    if value is not None:
+        if not REGION_PATTERN.match(value):
+            raise click.BadParameter("'{}'. Region must be a valid AWS region.".format(value))
+    return value
+
+
 region_option = click.option('--region', envvar='AWS_DEFAULT_REGION', metavar='AWS_REGION_ID',
-                             help='AWS region ID (e.g. eu-west-1)')
+                             help='AWS region ID (e.g. eu-west-1)', callback=validate_region)
 parameter_file_option = click.option('--parameter-file', help='Config file for params', metavar='PATH')
 output_option = click.option('-o', '--output', type=click.Choice(['text', 'json', 'tsv']), default='text',
                              help='Use alternative output format')
@@ -182,13 +196,7 @@ def watching(w: bool, watch: int):
 # and start with an alpha character. Maximum length of the name is 255 characters.
 STACK_NAME_PATTERN = re.compile(r'^[a-zA-Z][a-zA-Z0-9-]*$')
 VERSION_PATTERN = re.compile(r'^[a-zA-Z0-9]+$')
-
-
-def validate_version(ctx, param, value):
-    if not VERSION_PATTERN.match(value):
-        raise click.BadParameter('Version must satisfy regular expression pattern "[a-zA-Z0-9]+"')
-    return value
-
+REGION_PATTERN = re.compile(r'^[a-z]{2}-[a-z]+-[0-9]$')
 
 DEFINITION = DefinitionParamType()
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -21,7 +21,7 @@ def test_invalid_definition():
         with open('myapp.yaml', 'w') as fd:
             yaml.dump(data, fd)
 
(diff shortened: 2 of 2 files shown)


senza/cli.py
```diff
@@ -153,8 +153,28 @@ def convert(self, value, param, ctx):
...
+    return value
+
+
+def validate_region(ctx, param, value):
+    """Validate Click region param parameter.
+    Get the list of AWS regions and check the value is in it."""
+
+    if value is not None:
+        ec2 = boto3.client('ec2', 'eu-west-1')
+        ec2_regions = ec2.describe_regions()
```
hjacobs: this looks dangerously like an extra remote AWS API call....

...and it is indeed (checking the `botocore` source) :disappointed:

hjacobs:
:-1: for now because of the extra AWS API call

kenden:
@hjacobs What's wrong with an extra API call? Slowing things down? Harder to mock?

Could we do it the hardcoded way then?
```
if value not in [ 'eu-west-1', 'ap-southeast-1', 'ap-southeast-2', 'eu-central-1', 'ap-northeast-2', 'ap-northeast-1', 'us-east-1', 'sa-east-1', 'us-west-1', 'us-west-2' ]:
    raise click.BadParameter('invalid')
```

hjacobs:
@kenden I'm just wondering what the goal is?

* extra AWS API call is just adding unnecessary overhead/latency
* regex validation or enumeration of regions is cumbersome to maintain --- the old `boto` library had a hardcoded list of regions and we had to patch it to support the new Frankfurt region

Where do you see the problem of not having validation (first AWS call will show the error)? Do you mistype your region frequently? I would advice putting your region in `~/.aws/config` or having something like this in `.bashrc` to switch between two of them:

```bash
alias west='export AWS_DEFAULT_REGION=eu-west-1'
alias central='export AWS_DEFAULT_REGION=eu-central-1'

**Comment by hjacobs** [comment]
Maybe @jmcs and @rafaelcaricio have an opinion on this too?

**Comment by hjacobs** [comment]
@kenden I'm just trying to understand your use case --- I'm also (obviously) a frequent user of Senza, and I never missed the region validation....

**Comment by jmcs** [comment]
I personally prefer to follow a "better ask for forgiveness than permission" approach. The error handling needs to improved though.

File: senza/cli.py [hunk 2]
```diff
@@ -182,13 +197,7 @@ def watching(w: bool, watch: int):
...
 STACK_NAME_PATTERN = re.compile(r'^[a-zA-Z][a-zA-Z0-9-]*$')
 VERSION_PATTERN = re.compile(r'^[a-zA-Z0-9]+$')
-
-
-def validate_version(ctx, param, value):
-    if not VERSION_PATTERN.match(value):
-        raise click.BadParameter('Version must satisfy regular expression pattern "[a-zA-Z0-9]+"')
-    return value
-
+REGION_PATTERN = re.compile(r'^(eu|ap|us|sa)-(west|east|central|southeast|northeast)-(1|2)$')
```
rafaelcaricio: Maybe we could go with a more generic region regex instead of specific list? Maybe `^[a-z]{2}-[a-z]+-[0-9]$` would be enough? Seems like you just want to have a general validation to catch cases of really missing the parameter.

jmcs: The problem with this approach is that it will fail if Amazon adds an extra region to a place that already as 2 or some region in Africa for example.

rafaelcaricio: Not with this `^[a-z]{2}-[a-z]+-[0-9]$` though.

hjacobs:
@kenden yes, I agree that the "Unknown Error" message is not nice --- we had a stacktrace before which was not much better but at least it showed the cause without digging into another file :smirk: 

@rafaelcaricio @jmcs can we at least improve this "Unknown Error" handling?

@kenden is the regex validation proposal fine for you? How did you "mistype" the region when you got the error?

kenden:
@hjacobs I forgot the region altogether :) (check https://github.com/zalando-stups/senza/issues/233 for the stacktrace).
```
senza status --region senza-stack-template.yml stack1
```
@rafaelcaricio 
Using the more generic regex ``^[a-z]{2}-[a-z]+-[0-9]$`` would be fine, thanks!
I'll update and squash the commits.

rafaelcaricio:
I agree we should improve the generic error handing, but that is another issue and not related to this PR.

kenden:
@hjacobs I updated the regexp and squashed the commits.

hjacobs:
@kenden looks fine now --- just one last request: can you change the test code back to use some fictional AWS region ID? IMHO that's safer and prevents people from running into troubles someday where their test "works locally" because they have AWS credentials, but the test actually runs against real AWS :smirk: 

Fictional region ID makes sure that tests use mocks.

kenden:
@hjacobs Good point.I updated the test region in the tests to be 'aa-fakeregion-1'.
It passes the regex but won't work in AWS.

hjacobs:
:+1:

PR #239, opened 2016-06-17
Add better output for the user to command 'console'.

kenden:
- when the instance is not found
 - when the stack is not found
 - when there are no instances in the region.

In those cases ``senza console`` just returned and displayed nothing, which was confusing for the user.
Add usage examples to ``console --help``.

Fix and add tests.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1176,7 +1176,17 @@ def print_console(line: str):
 def console(instance_or_stack_ref, limit, region, w, watch):
     '''Print EC2 instance console output.
 
-    INSTANCE_OR_STACK_REF can be an instance ID, private IP address or stack name/version.'''
+    INSTANCE_OR_STACK_REF can be an instance ID, private IP address or stack name/version.
+
+    Ex:\n
+    senza console my-stack-template.yml version1\n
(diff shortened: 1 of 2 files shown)


hjacobs:
Thanks for your PR :smile:

hjacobs:
:+1:

senza/cli.py
```diff
@@ -1211,6 +1224,15 @@ def console(instance_or_stack_ref, limit, region, w, watch):
...
                         print_console(line)
+        if not found_match:
+            if stack_refs is None:
+                print("No EC2 instance with id '{}'.".format(instance_or_stack_ref[0]))
+            elif instance_or_stack_ref:
+                print("No stack matching '{}'{}.".format(
+                    stack_refs[0].name,
+                    " with version '{}'".format(stack_refs[0].version) if stack_refs[0].version is not None else ''))
+            else:
+                print("No EC2 instances in region '{}'.".format(region))
```
rafaelcaricio: Probably we should change the return code to 1 in all those cases...

kenden: @rafaelcaricio thanks. fix in https://github.com/zalando-stups/senza/pull/247

PR #241, opened 2016-06-21
Add flag to expose error instead of writing it to a file

### LINKED ISSUE #238: Add flag to expose error instead of writing it to a file
Would be nice to have a option to explicitly tell Senza to show the unknown exception stack trance in the console instead of writing it to a file.

### Current behaviour

It shows this message:
```
Unknown Error.
  Please create an issue with the content of /var/folders/90/pz80sxl93gl11zjsbmylwmxj42vx03/T/senza-traceback-_bv0iaac
```

### Expected behaviour
In case of calling Senza with something like `senza create broken.yaml --stacktrace-visible` then it would simply display the whole stack trace of the error in the console.

### Why?

It is important for tools like Lizzy that call Senza commands in background be able to pipe the actual errors that are happing while running Senza to the logs.

jmcs:
Fixes #238


senza/arguments.py
```diff
@@ -0,0 +1,48 @@
...
+
+json_output_option = click.option('-o', '--output',
+                                  type=click.Choice(['json', 'yaml']),
+                                  default='json',
+                                  help='Use alternative output format')
+
+stacktrace_visible_option = click.option('--stacktrace-visible',
+                                         is_flag=True,
+                                         help='Show stack trace instead of '
+                                              'storing it')
```
rafaelcaricio: Maybe it is a good idea to move the code:
```python
HandleExceptions.stacktrace_visible = stacktrace_visible
```
to something like a `callback=set_stacktrace_visible` here?

jmcs: Done.

rafaelcaricio:
👍

PR #245, opened 2016-06-21
Handle ClientError with code AccessDenied

kenden:
it is now handled the same as a NoCredentialsError Exception.

ClientError with code AccessDenied can happen when switching to another AWS role
(with ``aws sts assume-role --role-arn...`` for example).
If the token returned from that command expired, we get an accessdenied.

Small refactor to avoid code duplication.

### DIFF
--- a/senza/error_handling.py
+++ b/senza/error_handling.py
@@ -23,27 +23,47 @@ def store_exception(exception: Exception) -> str:
     return file_name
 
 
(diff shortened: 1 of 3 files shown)


jmcs:
:+1:

kenden:
Thanks for checking @jmcs. 
Could you maybe take take a second look? 
I added 2 tests and moved a couple around.

PR #249, opened 2016-06-22
Make the interface-facing ELB warning less conspicuous

### LINKED ISSUE #129: senza create: Don't give a false ararm
Every single time when I create a new *internet-facing* stack I receive an alarm: 
> You are deploying an internet-facing ELB that will be publicly accessible! You should have OAUTH2 and HTTPS in place!

**As** deploy engineer 
**I want to** suppress this warning into notification.
**So** that makes real errors  more visible and warnings stop [crying 'Wolf'](https://en.wikipedia.org/wiki/The_Boy_Who_Cried_Wolf)

jmcs:
Fixes #129


rafaelcaricio:
👍

PR #246, opened 2016-06-22
Support UPDATE_COMPLETE in Wait

jmcs:
Fixes #234

PR #251, opened 2016-06-22
Add testcase to test_wait. Remove wait --delete

kenden:
After the refactor in https://github.com/zalando-stups/senza/pull/246,
the option for senza wait ``--delete`` is not needed anymore.

Small refactor (especially renaming ``stacks_nok`` to ``stacks_in_progress`` improves readability a lot, IMHO).

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1430,17 +1430,27 @@ def wait(stack_ref, region, deletion, timeout, interval,):
 
     while time.time() < cutoff:
         stacks_ok = set()
-        stacks_nok = set()
+        stacks_in_progress = set()
         successful_actions = set()
-        for stack in get_stacks(stack_refs, region, all=True, unique_only=True):
+
+        stacks_found = get_stacks(stack_refs, region, all=True, unique_only=True)
+
+        if not stacks_found:
+            raise click.UsageError('No matching stack for "{}" found'.format(' '.join(stack_ref)))
+
+        for stack in stacks_found:
+
             if stack.StackStatus in target_status:
                 successful_action, _ = stack.StackStatus.split('_')
                 successful_actions.add('{}d'.format(successful_action.lower()))
                 stacks_ok.add((stack.name, stack.version))
-            elif stack.StackStatus.endswith('_FAILED') or stack.StackStatus.endswith('_COMPLETE'):
+
+            elif stack.StackStatus.endswith('_IN_PROGRESS'):
+                stacks_in_progress.add(stack.name, stack.version, stack.StackStatus)
+
+            else:  # _FAILED or ROLLBACK_COMPLETE or UPDATE_ROLLBACK_COMPLETE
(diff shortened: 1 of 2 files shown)


hjacobs:
What's the reason to remove `--deletion`?

kenden:
@hjacobs it's not needed and adds confusion.
I ran:
``
senza delete project-stack v1
``
then:
```
$ senza wait project-stack v1
Waiting up to 1800 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1795 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1790 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1785 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1779 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1773 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1654 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
Waiting up to 1648 more secs for stack bdf-minion-stacktest (DELETE_IN_PROGRESS)..
ERROR: Stack bdf-minion-stacktest has status DELETE_COMPLETE
```
and I thought it was a bug. 
I was just missing ``--delete``
Now there's no need to add this, wait works the same where creating, updating and deleting a stack.

hjacobs:
@kenden I think the main use case for `senza wait` is automated CI/CD pipelines (that's why I introduced it in the first place), so having an explicit flag for deletion is safer as you know "what" you want to wait for.

Also removing `--deletion` breaks our existing CD pipelines, so :-1: for that change from me.

kenden:
@hjacobs now that the wait command supports updates though, it's a bit weird to have a special case for deletion. Should it also have an `--update` param? That would be a bit much?

I re-added the --deletion param (I kept the extra test and small refactor).

senza/cli.py
```diff
@@ -1449,28 +1459,23 @@ def wait(stack_ref, region, deletion, timeout, interval,):
...
                                            for x in sorted(stacks_ok)])
             actions = '/'.join(successful_actions)
             ok('OK: Stack(s) {} {} successfully.'.format(successful_stacks,
                                                          actions))
             return
-        else:
-            raise click.UsageError('No matching stack for '
-                                   '"{}" found'.format(' '.join(stack_ref)))
-        time.sleep(interval)
-    raise click.Abort()
```
hjacobs: removing the `click.Abort` would result in exit code 0 if the timeout (`cutoff`) is reached

kenden:
@hjacobs thanks, fixed

hjacobs:
:+1:

hjacobs:
:+1:

PR #264, opened 2016-07-11
Show error when stack is not found

### LINKED ISSUE #263: Does not show error when stack does not exist
When I call:

```
$ senza delete this.stack.does.not.exists v33 --region eu-central-1
```

Senza does not show me an error that the stack does not exist neither returns 1. So if I have a typo in the name nothing happens at all leaving me without clue of what is happening.

rafaelcaricio:
Fixes #263

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -708,6 +708,9 @@ def delete(stack_ref, region, dry_run, force, interactive):
         fatal_error('Error: {} matching stacks found. '.format(len(stacks)) +
                     'Please use the "--force" flag if you really want to delete multiple stacks.')
 
+    if not stacks:
+        fatal_error('Error: Stack {} not found!'.format(stack_refs[0].name))
+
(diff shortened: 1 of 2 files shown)


jmcs:
Can you add an option to ignore if the stack doesn't exist (`-f` would be ideal to mirror `rm` but it's already taken :( )?

rafaelcaricio:
@jmcs in this case the error is only shown when there is not found stacks at all. So even without option if there is more than one stack to delete, it will ignore.

jmcs:
There can be situations (specially on CI\CD jobs) where I don't care if the stack exists or not, I just want to make sure it doesn't exist after I run the command.

rafaelcaricio:
@jmcs done!

jmcs:
:+1:

rafaelcaricio:
👍

Issue #266, opened 2016-07-12
LB Status not working for our setup


zmeda:
Hey!

We (one of the Zalando teams) have following setup. In `kio` we have registered single application called `example-app`. We want to support 3 different environments:
* integration
* staging
* production

In AWS we created 3 different security groups for each environment (also for load balancer). We have 3 different `yaml` files for each environment. Example of the `integration.yaml` config:

    - AppLoadBalancer:
      Type: Senza::WeightedDnsElasticLoadBalancer
      HTTPPort: 8080
      HealthCheckPath: /health
      SecurityGroups:
        - app-example-app-integration-lb
      Scheme: internet-facing
      MainDomain: "example-app-integration.team.zalan.do"
      LoadBalancerName: "example-app-integration-{{SenzaInfo.StackVersion}}"
      VersionDomain: "example-app-integration-{{SenzaInfo.StackVersion}}.team.zalan.do"

For all our created stack in AWS command `senza instances` shows no `LB Status` like:

    Stack Name           │Ver.          │Resource ID  │Instance ID        │Public IP│Private IP    │State  │LB Status │Launched
    example-app           integration1   AppServer     id-of-instance      x.w.z.y                  RUNNING              1d ago

As you can see, `LB Status` is not `IN_SERVICE` as expected. Some services are relying on this status so fix would be needed.

Thanks!

PR #269, opened 2016-07-19
#266 LB Status not working for our setup

meln1k:
p.s. can you provide docs to setup dev environment?

### DIFF
--- a/senza/components/elastic_load_balancer.py
+++ b/senza/components/elastic_load_balancer.py
@@ -116,7 +116,9 @@ def component_elastic_load_balancer(definition, configuration, args, info, force
                                               health_check_port,
                                               health_check_path)
 
-    if configuration.get('NameSuffix'):
+    if configuration.get('LoadBalancerName'):
+        loadbalancer_name = info["LoadBalancerName"]
+    elif configuration.get('NameSuffix'):
         version = '{}-{}'.format(info["StackVersion"],
                                  configuration['NameSuffix'])
         loadbalancer_name = get_load_balancer_name(info["StackName"], version)


senza/components/elastic_load_balancer.py
```diff
@@ -116,7 +116,9 @@ def component_elastic_load_balancer(definition, configuration, args, info, force
                                               health_check_port,
                                               health_check_path)
 
-    if configuration.get('NameSuffix'):
+    if configuration.get('LoadBalancerName'):
+        loadbalancer_name = info["LoadBalancerName"]
```
hjacobs: wrong reference

hjacobs:
Please add a test and describe your use case.

hjacobs:
I think your whole PR is obsolete as `LoadBalancerName` can already be overwritten (last part of the code sets any properties not defined by Senza).

senza/components/elastic_load_balancer.py
nehalium: If you are trying to resolve a load balancer name like "example-app-integration-{{SenzaInfo.StackVersion}}", then would you not rather do this? 
```
loadbalancer_name = get_load_balancer_name(info["LoadBalancerName"], info["StackVersion"])
```
Where you set LoadBalancerName to "example-app-integration".

zmeda: @hjacobs would it be correct?

nehalium: Of course, it would still be overwritten by the code block at the bottom. You would need to prevent this from happening somehow.

Issue #271, opened 2016-07-19
Some defaults are not sane for postgresapp


feikesteenbergen:
We should provide better defaults for Spilo:

- overcommit_ratio should be higher to run Spilo on smallish instances
- we should provide some sane logging defaults

will provide fix

Issue #270, opened 2016-07-19
zmon security group no longer recognized by postgresapp


feikesteenbergen:
The template for the postgresapp uses some assumptions to find the zmon security group.
With the newer zmon appliance, these assumptions are no longer valid.

Will provide fix

PR #272, opened 2016-07-19
Feature/spilo bugfixes

feikesteenbergen:
Addresses issues #271 and #270

### DIFF
--- a/senza/templates/postgresapp.py
+++ b/senza/templates/postgresapp.py
@@ -102,10 +102,12 @@ SenzaComponents:
           PGPASSWORD_STANDBY: "{{pgpassword_standby}}"
           BACKUP_SCHEDULE: "00 01 * * *"
           PATRONI_CONFIGURATION: | ## https://github.com/zalando/patroni#yaml-configuration
-            # postgresql:
-            #   pg_hba:
-            #   - hostnossl all all all reject
-            #   - hostssl   all all all md5
+            postgresql:
+            # pg_hba:
+            # - hostnossl all all all reject
+            # - hostssl   all all all md5
+              parameters:
+                log_connections: on
         root: True
         sysctl:
           vm.overcommit_memory: 2
@@ -243,7 +245,6 @@ Resources:
           - Effect: Allow
             Action:
               - "kms:Decrypt"
-              - "kms:Encrypt"
             Resource:
               - {{kms_arn}}
           {{/kms_arn}}
--- a/senza/templates/postgresapp.py
+++ b/senza/templates/postgresapp.py
@@ -109,7 +109,7 @@ SenzaComponents:
         root: True
         sysctl:
           vm.overcommit_memory: 2
-          vm.overcommit_ratio: 60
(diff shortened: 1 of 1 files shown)


senza/templates/postgresapp.py
```diff
@@ -243,7 +245,6 @@
           - Effect: Allow
             Action:
               - "kms:Decrypt"
-              - "kms:Encrypt"
```
jmcs: Why are you removing this? Why was it here before?

senza/templates/postgresapp.py
```diff
@@ -419,14 +420,14 @@ def gather_user_variables(variables, region, account_info):
...
-    filters = [{'Name': 'tag-key', 'Values': ['StackName']}, {'Name': 'tag-value', 'Values': ['zmon-worker']}]
+    filters = [{'Name': 'tag-key', 'Values': ['StackName']}, {'Name': 'tag-value', 'Values': ['zmon-appliance']}]
     zmon_sgs = list()
     for reservation in ec2.describe_instances(Filters=filters).get('Reservations', []):
         for instance in reservation.get('Instances', []):
             zmon_sgs += [sg['GroupId'] for sg in instance.get('SecurityGroups', []) if 'zmon' in sg['GroupName']]
 
     if len(zmon_sgs) == 0:
-        warning('Could not find zmon security group')
+        warning('Could not find zmon security group, do you have the zmon-appliance deployed?')
```
rafaelcaricio: In this case, since the user cannot answer this question. Maybe would be better to display "Could not find zmon security group, please verify if you have the zmon-appliance deployed!". What do you think?

senza/templates/postgresapp.py
feikesteenbergen: See commit dfafff128718dc032500c744e80033a850935af2, it was never used. Trying to apply the principle of least privilege here.

senza/templates/postgresapp.py
feikesteenbergen: They don't have to have zmon deployed, it's up to them. We do however want to try to find it *if* they have zmon deployed. I could change the wording, something like:

> Could not find zmon security group, skipping Security Group configuration for zmon. You can always add the zmon security group after deployment

senza/templates/postgresapp.py
```diff
@@ -102,14 +102,16 @@
           PGPASSWORD_STANDBY: "{{pgpassword_standby}}"
           BACKUP_SCHEDULE: "00 01 * * *"
           PATRONI_CONFIGURATION: | ## https://github.com/zalando/patroni#yaml-configuration
-            # postgresql:
-            #   pg_hba:
-            #   - hostnossl all all all reject
-            #   - hostssl   all all all md5
+            postgresql:
+            # pg_hba:
```
avaczi: Do we need these commented-out lines?

feikesteenbergen: No

hjacobs:
:+1:

Issue #273, opened 2016-07-21
metrictype is mandatory when autoscaling config is passed


drummerwolli:
use-case: i wrote my own custom metrics for my application which are send to cloudwatch. i want now to deploy my stack with this, but i can't set the upper and lower borders for the autoscaling group, since when i pass the autoscaling config, it requires me to have also a metric type (which must be CPU or network). But i want to have just the borders and then later one in the resources section define my own thresholds.

the code for this is here: https://github.com/zalando-stups/senza/blob/c66354e90048776c9b58b79946b3dc5cc82371f2/senza/components/auto_scaling_group.py#L182
basically i would say metrictype should be optional, to allow custom metrics.

workaround currently:
- use one of the existing metric type with unreachable thresholds (probably network)
- define the autoscaling group completely custom in the resources section
(both of these are super ugly)

PR #274, opened 2016-07-21
make metrictype optional

drummerwolli:
fixes https://github.com/zalando-stups/senza/issues/273


hjacobs:
@drummerwolli can you add a test for this case?

drummerwolli:
will adapt as well in the afternoon the documentation ...

hjacobs:
:+1:

PR #275, opened 2016-07-22
Support and use alias records

### LINKED ISSUE #197: Use Alias Route 53 records instead of CNAMEs
http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html?tag=duckduckgo-d-20

Why change from CNAME to Alias records?
* Alias records allow you to use Senza with your “root” team domain (e.g. you can create a stack serving “myteam.example.org”)
* Alias records allow having TXT or MX records with the same domain name
* Alias records avoid one DNS lookup (clients need to lookup CNAME, afterwards resolve the target to the IP)
* You can optionally enable “Evaluate Target Health” to have a DNS failover (AWS will only direct traffic to the Alias target if ELB health is OK)

jmcs:
Fixes #197.


jmcs:
For whatever reason doctests start making py.test go on an infinite loop, so I'll reimplement the tests as "proper" unit tests.

.coveragerc
```diff
@@ -2,3 +2,5 @@
 exclude_lines =
     pragma: no cover
     def __repr__
+omit =
+    senza/__main__.py
```
rafaelcaricio: Missing NEWLINE.

senza/components/elastic_load_balancer.py
```diff
@@ -31,18 +32,19 @@ def component_elastic_load_balancer(definition, configuration, args, info, force
...
-            },
-        }
+
+        domain_name = "{0}.{1}".format(domain["Subdomain"], domain["Zone"])
+
+        convert_domain_records_to_alias(domain_name)
+
+        properties = {"Type": "A",
+                      "Name": domain_name,
+                      "HostedZoneName": "{0}".format(domain["Zone"]),
```
rafaelcaricio: Maybe `"HostedZoneName": domain["Zone"],` ?

senza/manaus/cloudformation.py
```diff
@@ -0,0 +1,154 @@
...
+                 status: str,
+                 stack_status_reason: Optional[str],
+                 disable_rollback: bool,
+                 notification_arns: List[str],
+                 timeout_in_minutes: Optional[int],
+                 capabilities: Optional[List[str]],
+                 outputs: Optional[List[Dict]],
+                 tags: Dict[str, str],
+                 *,
+                 region: Optional[str]):
```
rafaelcaricio: Maybe move this up? Before the `*`.

jmcs: I don't want region set by accident, since it's not returned by AWS, it's useful information but comes from the HostedZone when searching for domains inside it.

.coveragerc
jmcs: Fixed.

senza/manaus/constants.py
```diff
@@ -0,0 +1,11 @@
+ELB_REGION_HOSTED_ZONE = {"us-east-1": "Z35SXDOTRQ7X7K",
```
rafaelcaricio: Didn't we agreed in [another PR](https://github.com/zalando-stups/senza/pull/237) to not have a list of AWS zones?

jmcs: This is far from ideal, but I couldn't find a good way to get the hosted zone of the ELBs for an arbitrary region, I only found the table in a documentation page (I should probably leave a comment here with a link to the documentation)

hjacobs: Why do we have hardcoded hosted zone IDs here? Is there no better way? This looks brittle...

senza/manaus/elb.py
```diff
@@ -0,0 +1,24 @@
...
+                            region: Optional[str]=None) -> Route53HostedZone:
+
+        if all([dns_name, region]):
+            raise ValueError("Provide only one of dns_name and region.")
+
+        if dns_name is not None:
+            # dns_name = app-name.REGION.elb.amazonaws.com
+            _, region, _ = dns_name.split('.', maxsplit=2)
+
+        hosted_zone_id = ELB_REGION_HOSTED_ZONE[region]
```
rafaelcaricio: We should use `.get` with a default or handle somehow a possible `KeyError` in case of a hosted zone that is not present in the list.

senza/manaus/route53.py
```diff
@@ -35,12 +60,61 @@ def from_boto_dict(cls, hosted_zone_dict: Dict[str, Any]) -> 'Route53HostedZone'
...
+        http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.change_resource_record_sets
+        """
+
+        client = boto3.client('route53')
+
+        change_batch = {'Changes': []}
+        if comment is not None:
+            change_batch['Comment'] = comment
+        for record in changed_records:
+            change = {'Action': 'UPSERT',
```
rafaelcaricio: Maybe we should extract this method behaviour? It is duplicated (like `delete`) and the only difference is the 'UPSERT' and 'DELETE' values in the "change" `dict`.

senza/manaus/constants.py
jmcs: I think I found a way to get the hosted zone in runtime. I'm going to try to implement it.

tests/test_utils.py
```diff
@@ -0,0 +1,5 @@
+from senza.utils import camel_case_to_underscore
+
+
+def test_camel_case_to_underscore():
+    assert camel_case_to_underscore('CamelCaseToUnderscore')  == 'camel_case_to_underscore'
```
rafaelcaricio: NEWLINE missing.

senza/manaus/route53.py
jmcs: I'll do that

jmcs:
I added region to all clients and resources to avoid problems like #281.

senza/manaus/route53.py
```diff
@@ -35,12 +69,80 @@ def from_boto_dict(cls, hosted_zone_dict: Dict[str, Any]) -> 'Route53HostedZone'
...
+        """
+        Submits changes to be upsterted in Route53. Returns the dict sent to
+        AWS.
+
+        See:
+        http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.change_resource_record_sets
+        """
+
+        changes = [(ChangeAction.UPSERT, record) for record in changed_records]
+
```
rafaelcaricio: Why so many blank lines all around those functions/classes?

rafaelcaricio:
👍

hjacobs:
This is a huge PR and has many unrelated changes... I fear we will introduce problems here...

hjacobs:
@jmcs what is the migration path? What happens if I have a stack deployed with old Senza and deploy a new stack version with new Senza?

jmcs:
The user is asked to convert the existing records, most changes done on this pull request are to support that conversion.

hjacobs:
@jmcs deploying a new stack (all stack versions were deleted) fails if dangling DNS records exist:

```
$ python3 -m senza create ../foo/zmon-controller-alpha.yaml cd387c1 cd387
Generating Cloud Formation template..You are deploying an image that has *MEDIUM* severity security fixes easily available!  Please check this artifact tag in pierone and see which software versions you should upgrade to apply those fixes.
 EXCEPTION OCCURRED: ELB not found: alpha-cd385c3-123.eu-central-1.elb.amazonaws.com
ELB not found: alpha-cd385c3-123.eu-central-1.elb.amazonaws.com

jmcs:
@hjacobs How do you propose solving that case? Deleting the dangling record?

hjacobs:
@jmcs also another "problem": `senza status ..` will not show `main_dns` correctly anymore (it's based on CNAME lookup)

hjacobs:
@jmcs maybe documentation is enough, we should document all pitfalls and how to fix them (e.g. here by deleting the dangling record), see https://github.com/zalando-stups/senza/issues/197#issuecomment-236558247 .

hjacobs:
@jmcs this behavior might be problematic for test stacks which are regularly deleted (production stacks usually have at least one version running..)

PR #284, opened 2016-08-01
Fix some issues caused by #275

jmcs:
- Fix Main DNS in senza status;
- Delete dangling records when converting to Alias records;

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -952,9 +952,13 @@ def status(stack_ref, region, output, w, watch):
                         except:
                             http_status = 'ERROR'
                     else:
-                        try:
-                            answers = dns.resolver.query(name, 'CNAME')
-                        except:
+                        for record_type in ["CNAME", "A", "AAAA"]:
+                            try:
+                                answers = dns.resolver.query(name, record_type)
+                                break
+                            except:
+                                pass
+                        else:
                             answers = []
                         for answer in answers:
                             target = answer.target.to_text()
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -952,13 +952,9 @@ def status(stack_ref, region, output, w, watch):
                         except:
                             http_status = 'ERROR'
                     else:
-                        for record_type in ["CNAME", "A", "AAAA"]:
-                            try:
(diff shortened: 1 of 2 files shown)


senza/cli.py
```diff
@@ -952,9 +952,13 @@ def status(stack_ref, region, output, w, watch):
...
                             http_status = 'ERROR'
                     else:
-                        try:
-                            answers = dns.resolver.query(name, 'CNAME')
-                        except:
+                        for record_type in ["CNAME", "A", "AAAA"]:
+                            try:
+                                answers = dns.resolver.query(name, record_type)
+                                break
+                            except:
```
rafaelcaricio: Really any error?

senza/cli.py
```diff
@@ -952,9 +952,13 @@ def status(stack_ref, region, output, w, watch):
                         except:
                             http_status = 'ERROR'
                     else:
-                        try:
-                            answers = dns.resolver.query(name, 'CNAME')
-                        except:
+                        for record_type in ["CNAME", "A", "AAAA"]:
```
hjacobs: but this will still not work as the `answer.target.to_text()` will not contain the stack version (only an IP..)

jmcs: @hjacobs I reverted that change and I'll create a Pull Request later to fix this (maybe by checking the route53 api).

senza/manaus/route53.py
```diff
@@ -337,9 +337,12 @@ def convert_domain_records_to_alias(domain_name: str):
...
     for record in records:
         if record.type != 'A':
-            alias_record = record.to_alias()
             converted_records[record.hosted_zone]['delete'].append(record)
-            converted_records[record.hosted_zone]['upsert'].append(alias_record)
+            try:
+                alias_record = record.to_alias()
+                converted_records[record.hosted_zone]['upsert'].append(alias_record)
+            except ELBNotFound:
+                print("Dangling DNS record  will be deleted: {}".format(record.name))
```
hjacobs: extra space between "record" and "will"

hjacobs:
:+1:

PR #285, opened 2016-08-02
Fix senza status for alias records

### LINKED ISSUE #197: Use Alias Route 53 records instead of CNAMEs

hjacobs:
Fixes #197

### DIFF
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -1,4 +1,5 @@
 import collections
+import dns.resolver
 from json import JSONEncoder
 
 import boto3
@@ -395,3 +396,12 @@ def inform_sns(arns: list, message: str, region):
     sns = boto3.client('sns', region_name=region)
     for sns_topic in sns_topics:
         sns.publish(TopicArn=sns_topic, Subject="SenzaTrafficRedirect", Message=jsonizer.encode((message)))
+
+
+def resolve_to_ip_addresses(dns_name: str) -> set:
+    try:
+        answers = dns.resolver.query(dns_name, 'A')
+    except:
+        return set()
+    else:
+        return {answer.address for answer in answers}
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -399,6 +399,9 @@ def inform_sns(arns: list, message: str, region):
 
 
 def resolve_to_ip_addresses(dns_name: str) -> set:
+    """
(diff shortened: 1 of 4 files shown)


senza/traffic.py
```diff
@@ -395,3 +396,12 @@ def inform_sns(arns: list, message: str, region):
     sns = boto3.client('sns', region_name=region)
     for sns_topic in sns_topics:
         sns.publish(TopicArn=sns_topic, Subject="SenzaTrafficRedirect", Message=jsonizer.encode((message)))
+
+
+def resolve_to_ip_addresses(dns_name: str) -> set:
+    try:
+        answers = dns.resolver.query(dns_name, 'A')
+    except:
```
rafaelcaricio: It really bothers me those "god mode" :godmode: exceptions catchers. Maybe we can try to be a bit more specific? Maybe [DNSException](https://github.com/rthalley/dnspython/blob/master/dns/exception.py#L19)?

hjacobs: @rafaelcaricio I personally don't care if the contents of the `try` clause only have one call to a 3rd party library --- I don't want to look up what kind of exceptions the 3rd party raises, I just want to ignore any error..

hjacobs: E.g. maybe the 3rd party library could raise some IO or OS errors and I did not thought of that...

hjacobs:
:+1:

senza/traffic.py
rafaelcaricio: If the 3rd party call gives a IO or OS error I would probably care about to know what is going on there since my code is somehow depending on the 3rd library. If the call is not important or if the library does not work, I would probably prefer to remove the dependency.

Even in one line the error can be from our side -- maybe we are passing a wrong value to the library and ignoring the error that would give me how to solve the issue. Silencing everything, in my opinion, does not sound good.

PR #291, opened 2016-08-03
Target health check

### LINKED ISSUE #286: Consider enabling "Evaluate Target Health" for Alias DNS records
To get a nice DNS failover (only direct traffic to stacks with healthy ELB).

jmcs:
Fixes #286 

Note: I still didn't test this changes with real stacks

### DIFF
--- a/senza/manaus/route53.py
+++ b/senza/manaus/route53.py
@@ -284,7 +284,7 @@ class Route53Record:
 
             alias_target = {"HostedZoneId": hosted_zone_id,
                             "DNSName": dns_name,
-                            "EvaluateTargetHealth": False}
+                            "EvaluateTargetHealth": True}
             return self.__class__(name=self.name,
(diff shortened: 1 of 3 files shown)


rafaelcaricio:
👍

hjacobs:
I would not make this the default --- having this always on might lead to undesirable behavior (DNS lookup failure instead of "Service Temporarily Not Available").

I would leave it to the user whether he wants DNS failover (which requires having at least one other stack deployed).

We should also test the actual AWS behavior (how does it behave if the other stack has weight zero etc?).

hjacobs:
Hmm, actually we delete DNS records with weight zero, so DNS failover would not work out-of-the-box with one "primary" (weight 200) and one "secondary" (weight 0) stack.

jmcs:
I'm going to close the PR until I can do some tests.

Issue #299, opened 2016-08-08
postgresapp (Spilo) template raises exception with empty scalyr key


feikesteenbergen:
Scalyr key should only be encrypted and injected if it is not empty. Currently when empty

```text
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the Encrypt operation: 1 validation error detected: Value at 'plaintext' failed to satisfy constraint: Member must have length greater than or equal to 1
```

PR #300, opened 2016-08-08
Ensure variables are not empty when encrypting

feikesteenbergen:
To address issue #299, we should only encrypt non-empty
variables.


feikesteenbergen:
👍

alexeyklyukin:
👍

hjacobs:
:+1:

PR #302, opened 2016-08-09
Get record by name

jmcs:
This method was initially done has part of #275 but ended up nothing being used. It can be useful in the future though.


rafaelcaricio:
👍

PR #304, opened 2016-08-09
Only call API for Route53 config if there is changes to be made

### LINKED ISSUE #303: Only call API for Route53 config if there is changes to be made
Only call AWS API when there is actual changes to be made.

Users reporting this exception:
```
     raise ParamValidationError(report=report.generate_report())
 botocore.exceptions.ParamValidationError: Parameter validation failed:
 Invalid length for parameter ChangeBatch.Changes, value: 0, valid range: 1-inf
```

Error comes from https://github.com/zalando-stups/senza/blob/master/senza/manaus/route53.py#L114-L115

rafaelcaricio:
Fixes #303

### DIFF
--- a/senza/manaus/route53.py
+++ b/senza/manaus/route53.py
@@ -112,8 +112,9 @@ class Route53HostedZone:
                       'ResourceRecordSet': record.boto_dict}
             change_batch['Changes'].append(change)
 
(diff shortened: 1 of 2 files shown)


hjacobs:
:+1:

PR #305, opened 2016-08-09
#37 allow setting the VpcId for StupsAutoConfiguration

### LINKED ISSUE #37: Support multiple VPCs
Senza's STUPS components currently assume a single VPC per region (Seven Seconds only configures a single VPC). There might be valid reasons to have multiple VPCs (e.g. a special VPC with VPN tunnel to some 3rd party location), so Senza needs to support that, too.

hjacobs:
Fixes #37 by allowing to set the VpcId in the Senza definition thus allowing multiple VPCs and selecting the right one manually.


tuxlife:
:+1:

rafaelcaricio:
👍

PR #309, opened 2016-08-11
Fix infinite loop calling AWS API when using senza wait

### LINKED ISSUE #308: When stacks does not exist senza wait gets in infinite loop
When using `senza wait` with a stack that does not exist, senza command does not show anything but keeps calling AWS API in a infinite loop.

rafaelcaricio:
Fixes #308

Problem was that generators are truthy so when stack were not found Senza was getting in an infinite loop.


jmcs:
:+1:

PR #313, opened 2016-08-15
Update Patroni config section to >= Patroni 1.0

feikesteenbergen:
In Patroni 1.0 a change of the configuration was introduced.
This change ensures newly created Spilo's default to Patroni 1.0
valid configuration.
Addresses https://github.com/zalando/spilo/issues/83


rafaelcaricio:
👍

PR #322, opened 2016-08-25
Find the ELB for traffic switch by dns name

### LINKED ISSUE #321: Traffic switching by CF update does not work for multiple domains or custom resource names
As far as I can see #318 introduces a bunch of problems:

* traffic switching is slow (waiting for AWS CloudFormation...)
* traffic switching does not work with custom resource names ("AppLoadBalancerMainDomain" hardcoded)
* traffic switching does not work with multiple load balancers / domain names (we have this scenario e.g. in our Plan B setup where we have different SSL certs and therefore different ELBs and DNS names)

jmcs:
Fixes #321

### DIFF
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -8,7 +8,7 @@ import dns.resolver
 from clickclick import Action, action, ok, print_table, warning
 
 from .aws import StackReference, get_stacks, get_tag
-from .manaus.cloudformation import CloudFormationStack
+from .manaus.cloudformation import CloudFormationStack, ResourceType
 from .manaus.exceptions import StackNotFound, StackNotUpdated
 from .manaus.route53 import (RecordType, Route53, Route53HostedZone,
                              convert_domain_records_to_alias)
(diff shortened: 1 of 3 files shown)


senza/traffic.py
```diff
@@ -150,8 +150,18 @@ def set_new_weights(dns_names: list, identifier, lb_dns_name: str,
                                                "because traffic for it is 0".format(stack_name))
                 changed = True
                 continue
-            load_balancer = stack.template['Resources']['AppLoadBalancerMainDomain']
-            load_balancer['Properties']['Weight'] = percentage
+
+            for key, resource in stack.template['Resources'].items():
+                if (resource['Type'] == ResourceType.route53_record_set and
+                        resource['Properties']['Name'] == dns_name):
+                    load_balancer = stack.template['Resources'][key]
```
hjacobs: can you change the variable name? `load_balancer` is misleading (as it's a Route53 record!)

PR #323, opened 2016-08-25
Don't connect to instances automatically if there's more than one

### LINKED ISSUE #311: Please set the --connect option as optional
After 479e4d8 you connect automatically to ***all*** instances if you use `--piu`

But this is not useful, if you get more then one instance back.

jmcs:
Fixes #311


hjacobs:
:+1:

PR #324, opened 2016-08-25
#321 fallback to direct DNS upsert if CF is not updated

hjacobs:
#321: fallback to direct manipulation if CloudFormation says "no update needed".

This is important for a seamless migration to the new Senza version (otherwise old versions (manipulated with old Senza) will still get DNS traffic!),


jmcs:
:+1:

PR #327, opened 2016-08-26
Sentry

### LINKED ISSUE #325: Add Sentry support
Senza should have optional sentry support. This would allow real-time exception tracking and consequently a faster response time to fix bugs.

jmcs:
Fixes #325 .

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -33,7 +33,9 @@ from .aws import (StackReference, get_account_alias, get_account_id,
                   parse_time, resolve_topic_arn)
 from .components import evaluate_template, get_component
 from .components.stups_auto_configuration import find_taupage_image
+from .exceptions import InvalidDefinition
 from .error_handling import HandleExceptions
+from .manaus.cloudformation import CloudFormation
 from .manaus.ec2 import EC2
 from .manaus.exceptions import VPCError
 from .manaus.route53 import Route53, Route53Record
@@ -73,6 +75,7 @@ STYLES = {
     'UPDATE_FAILED': {'fg': 'red'},
     'UPDATE_ROLLBACK_COMPLETE': {'fg': 'red'},
     'IN_SERVICE': {'fg': 'green'},
+    'HEALTHY': {'fg': 'green'},
     'OUT_OF_SERVICE': {'fg': 'red'},
     'OK': {'fg': 'green'},
     'ERROR': {'fg': 'red'},
@@ -133,7 +136,11 @@ class DefinitionParamType(click.ParamType):
                 #     url = 'file://{}'.format(quote(os.path.abspath(value)))
 
                 response = urlopen(url)
-                data = yaml.safe_load(response.read())
+                try:
+                    data = yaml.safe_load(response.read())
+                except yaml.YAMLError as e:
+                    raise InvalidDefinition(path=value,
+                                            reason=str(e))
             except URLError:
                 self.fail('"{}" not found'.format(value), param, ctx)
         else:
@@ -441,8 +448,21 @@ def get_stack_refs(refs: list):
         else:
             try:
                 with open(ref) as fd:
-                    data = yaml.safe_load(fd)
-                ref = data['SenzaInfo']['StackName']
+                    try:
+                        data = yaml.safe_load(fd)
+                    except yaml.YAMLError as e:
+                        raise InvalidDefinition(path=ref,
+                                                reason=str(e))
+
+                try:
+                    ref = data['SenzaInfo']['StackName']
+                except KeyError:
+                    raise InvalidDefinition(path=ref,
+                                            reason="SenzaInfo is missing "
+                                                   "or invalid")
+                except TypeError:
+                    raise InvalidDefinition(path=ref,
+                                            reason="Invalid SenzaInfo")
             except (OSError, IOError):
                 # It's still possible that the ref is a regex
                 pass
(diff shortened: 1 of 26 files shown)


rafaelcaricio:
@jmcs could you please describe what is this PR about?

senza/configuration.py
```diff
@@ -0,0 +1,66 @@
...
+        return section, sub_key
+
+    def __save(self, cfg):
+        self.config_path.parent.mkdir(parents=True, exist_ok=True)
+        with self.config_path.open('w+') as config_file:
+            yaml.safe_dump(cfg, config_file,
+                           default_flow_style=False)
+
+    @property
+    def dict(self) -> Dict[str, Dict[str, str]]:
```
rafaelcaricio: Maybe we could use a better naming here? `dict` is very misleading.

senza/configuration.py
```diff
@@ -0,0 +1,66 @@
...
+        try:
+            with self.config_path.open() as config_file:
+                cfg = yaml.safe_load(config_file)
+        except FileNotFoundError:
+            cfg = {}
+        return cfg
+
+configuration = Configuration()
```
rafaelcaricio: I would have this as `Configuration(Path(get_app_dir('senza')) / "config.yaml")` and leave the `Configuration` class clean of specifics.

senza/error_handling.py
```diff
@@ -116,3 +127,20 @@ def __call__(self, *args, **kwargs):
         except Exception as e:
             # Catch All
             self.die_unknown_error(e)
+
+
+# Setup Sentry
```
rafaelcaricio: :hocho:  useless comment.

senza/configuration.py
rafaelcaricio: It also would be much easier to test.

senza/configuration.py
```diff
@@ -0,0 +1,76 @@
...
+from pathlib import Path
+from typing import Dict, Tuple
+
+import yaml
+from click import get_app_dir
+
+from .exceptions import InvalidConfigKey
+
+
+class Configuration(MutableMapping):
```
hjacobs: this whole class looks complicated for what it does: load/store a dict as YAML

Also we already use the very naive approach with `stups_cli.load_config` in all the STUPS CLI tools.

jmcs: I wanted to support sections because I wanted both to stay consistent with git and allow extensions\templates to have their own configuration. The stups_cli config is also dangerous for config commands since it loads configuration from the environment (which is a nice feature) but then dumps the whole configuration to the disk when saving, which would mean that users could accidentally overwrite their configuration with values that they intended to use only temporarily.

senza/error_handling.py
rafaelcaricio: This comment is exactly the same text of the function it is commenting about. That kind of comment is just annoying. :disappointed:

senza/cli.py
```diff
@@ -1634,6 +1616,8 @@ def wait(stack_ref, region, deletion, timeout, interval):
             return
     raise click.Abort()
 
+cli.add_command(cmd_config)
```
rafaelcaricio: Is it really necessary to add `cmd_config` twice? It seems added [here](https://github.com/zalando-stups/senza/pull/327/files#diff-a8d4b26c8d4f994b3beedd0f80a63485R10) again.

tests/test_configuration.py
```diff
@@ -0,0 +1,64 @@
+from mock import MagicMock, mock_open
```
rafaelcaricio: We could use `from unittest.mock import MagicMock, mock_open`, no?

setup.py
```diff
@@ -130,7 +130,7 @@ def setup_package():
         install_requires=install_reqs,
         setup_requires=['flake8'],
         cmdclass=cmdclass,
-        tests_require=['pytest-cov', 'pytest'],
+        tests_require=['pytest-cov', 'pytest', 'mock'],
```
rafaelcaricio: Already included in `unittest.mock`.

tests/test_subcommands/test_config.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def setitem(key, value):
+        if '.' not in key:
+            raise InvalidConfigKey("Key Error")
+        mock.dict[key] = value
+
+    mock.__getitem__.side_effect = get
+    mock.__setitem__.side_effect = setitem
+
+    monkeypatch.setattr('senza.subcommands.config.configuration', mock)
```
rafaelcaricio: Wouldn't it be better to use the original class in the tests?

setup.py
jmcs: This is\was needed because it backports some methods to python 3.4

tests/test_configuration.py
jmcs: From https://docs.python.org/3.5/library/unittest.mock.html#unittest.mock.mock_open:

> Changed in version 3.5: read_data is now reset on each call to the mock.

We need this otherwise a second read would fail.

setup.py
jmcs: See the comment to test_configuration. Without this requirement the tests fail in python 3.4.

senza/cli.py
jmcs: Yes, because [@command](https://github.com/zalando-stups/senza/pull/327/files#diff-a8d4b26c8d4f994b3beedd0f80a63485R10) only makes the function a command. `add_command` makes it a subcommand of another command.

tests/test_subcommands/test_config.py
jmcs: Done

rafaelcaricio:
Tests are failing...

rafaelcaricio:
:+1:

rafaelcaricio:
tests still failing though.

PR #328, opened 2016-08-26
Introduce a support for CloudWatch metrics for AutoScaling groups

alterrebe:
One need to specify:

      CloudWatchMetrics: true

in the AutoScalingGroup or TaupageAutoScalingGroup to enable CloudWatch metrics collection.

### DIFF
--- a/senza/components/auto_scaling_group.py
+++ b/senza/components/auto_scaling_group.py
@@ -128,6 +128,12 @@ def component_auto_scaling_group(definition, configuration, args, info, force, a
         }
     }
 
+    if "CloudWatchMetrics" in configuration and configuration["CloudWatchMetrics"]:
+        # add collection of all ASG metrics once per a minute
+        definition["Resources"][asg_name]["Properties"]["MetricsCollection"] = [
+            {"Granularity": "1Minute"}
+        ]
+
     asg_properties = definition["Resources"][asg_name]["Properties"]
 
     if "OperatorTopicId" in info:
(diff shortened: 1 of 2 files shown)


hjacobs:
Aren't CloudWatch metrics already collected for ASGs?

alterrebe:
No, it is off by default for AS groups. You need to specify the MetricsCollection property in the CF template or switch it on in the AWS console afterwards.

hjacobs:
OK, but I rather would like to introduce a generic property override for AutoScalingGroup (similar to what we have with ELB) instead of another name mapping (`CloudWatchMetrics` -> `MetricsCollection`).

alterrebe:
A good idea, but it introduces some complexities:
* We need to avoid ambiguity - e.g. a developer can specify `MaxSize` that would be conflicting with `AutoScaling/Maximum` setting;
* We need to avoid name clashing. Currently there is no collisions between our properties and CF ones, but both things are evolving.

I would propose to put the CF options under a specific property, we can name it `Override` or `CFOptions`. Also Senza should check if few specific options (`MinSize`, `MaxSize` and few others) appear in the section and it should raise an error if they do.

Any comments?

hjacobs:
I would not introduce a new "Override" or "CF*" property, but just rely not having name collisions --- future name collisions might always be mitigated by introducing a new Senza component.

Senza components should be just a convenient wrapper for CF resources, using the same properties makes it IMHO simpler.

hjacobs:
I'll quickly do the necessary changes.

hjacobs:
@alterrebe please check #331

Issue #310, opened 2016-08-11
Support for Application Load Balancers


arjunrn:
AWS just released a new Load Balancer [type](https://aws.amazon.com/blogs/aws/new-aws-application-load-balancer/) with support for HTTP2. There is already support for it in Cloudformation. We could have a new component likewise called `Senza::WeightedDnsApplicationLoadBalancer`.

prayerslayer:
Properties of the ELB have been split up in multiple components for the ALB, and some were added.

* [LoadBalancer](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html): Basically `scheme`, `securityGroup`, `subnet`.
* [Listener](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html): `port`, `protocol`, `certificates`
* [ListenerRule](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html): Use this to forward requests to target groups based on the request path
* [TargetGroup](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html): Describes a set of EC2s along with health check config, VPC and such.

The ALB is awesome* as we could do something like this(?):

* Create ALB once with DNS record pointing to it, in separate CF stack
* Application stack contains TargetGroup for the ALB which points to its AutoScaling group (hope this works)
* Application stack also contains ListenerRule for the ALB stating when to route requests to it (e.g. for requests to `/api/*`

:scream:

* Somehow you can't route based on headers yet.

arjunrn:
It might be easier to just have another component in the same stack. I don't think many microservices will have a requirement for path based routing since that would be handled in something like Skipper anyway. So just use a match all wildcard like `/*` and use it exactly like the Classic LB. Maybe @hjacobs or @tuxlife could pitch in as well with their opinions.

jmcs:
I agree that it's better to have another component to avoid breaking things that are working now.

hjacobs:
We should probably introduce a new component, but AWS guys advice us to switch to ALB for everything  HTTP-based (some under-the-hood performance improvements etc).

So I would say:

* introduce new component (with "V2" suffix)
* switch default `senza init` templates to use the new component

[...]

PR #329, opened 2016-08-29
#310 code cleanup and allow overwritting ELBv2 properties

hjacobs:
#310: Further steps towards full ELBv2 support

* remove duplicate code
* allow overwriting ELBv2 / target group properties
* default health check path is now '/health' (much more reasonable, "/ui/" was just convenient for original STUPS microservices)


jmcs:
:+1:

PR #332, opened 2016-08-29
#330 truncate default CF description to 1024 chars

### LINKED ISSUE #330: Shrink "description" field to allowed 1024 chars
This seems to pop up more and more, I see no real reason why senza should not just do

description[:1024]

to fix this for everyone.

hjacobs:
Fixes #330


hjacobs:
:+1:

PR #339, opened 2016-09-01
Internal default scheme

### LINKED ISSUE #338: ELB should have Scheme "internal" if "Scheme" is not defined
The ELB "Scheme" property should default to "internal" if it's not set.

hjacobs:
Fixes #338.


hjacobs:
:+1:

PR #345, opened 2016-09-06
Fix bad message credential access error

kenden:
When getting the error: access_denied_error (which means: credentials are present in .aws/credentials or in env variables, but the associated user does not have proper access rights on the AWS resource),

Senza reported:
No AWS credentials found. Use the "mai" command-line tool to get a temporary access key
or manually configure either ~/.aws/credentials or WS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY.

Which is not correct: the credentials ARE present. That cost me a bit of time to find the issue.

Also,
- DRY up the code: do not repeat exit(1) all the time.
- Display errors using the clickclick.fatal_error function, which colors the error in red all the time.

### DIFF
--- a/senza/error_handling.py
+++ b/senza/error_handling.py
@@ -95,7 +95,7 @@ class HandleExceptions:
             err_mesg = "Error parsing definition file:\n{}".format(e)
             if e.problem == "found unhashable key":
                 err_mesg += "Please quote all variable values"
-                die_fatal_error(err_mesg)
+            die_fatal_error(err_mesg)
         except PiuNotFound as e:
             die_fatal_error(
                 "{}\nYou can install piu with the following command:"
--- a/senza/error_handling.py
+++ b/senza/error_handling.py
@@ -87,7 +87,7 @@ class HandleExceptions:
                         e.response['Error']['Message']))
             elif is_validation_error(e):
                 die_fatal_error(
-                    "Validation Error: \n{}".format(
+                    "Validation Error: {}".format(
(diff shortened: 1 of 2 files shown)


kenden:
I can't see the changes in Coverage in Coveralls:
https://coveralls.io/builds/7763828/source?filename=senza%2Ferror_handling.py

SOURCE NOT AVAILABLE
The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.

senza/error_handling.py
```diff
@@ -56,64 +61,50 @@ def __init__(self, function):
...
                 self.die_unknown_error(e)
         except yaml.constructor.ConstructorError as e:
-            print("Error parsing definition file:", file=sys.stderr)
-            print(e, file=sys.stderr)
+            err_mesg = "Error parsing definition file:\n{}".format(e)
             if e.problem == "found unhashable key":
-                print("Please quote all variable values", file=sys.stderr)
-            sys.exit(1)
+                err_mesg += "Please quote all variable values"
+                die_fatal_error(err_mesg)
```
hjacobs: indentation is wrong here

hjacobs:
@kenden thanks for the PR. I think the indentation is mixed up at one place (see my code comment).

kenden:
@hjacobs good catch! I fixed it.
That means there is coverage missing for the case
` if e.problem != "found unhashable key":`

jmcs:
:+1:

PR #347, opened 2016-09-07
Safe(r) delete

### LINKED ISSUE #292: Add option --fail-on-traffic to delete command
A suggestion:

senza delete --fail-on-traffic mystack 123 should cause senza to check whether there is traffic directed to the stack. If there is traffic directed to the stack, the command fails.

We would always use this option (we delete stacks via Jenkins job) and would feel more comfortable with this option.

### LINKED ISSUE #343: Fail on deletion of a stack with traffic
Current behaviour of senza when deleting a stack with 100% traffic is to silently remove the stack and DNS entries, which leads to "unknown host" for all users. This is potentially error-prone and can lead to downtimes in production. It would be better if it does one of the following

* either die and not allow to delete an instance that has (full / any) traffic
* or route traffic to one of the other stacks that already have traffic

Imho, the first proposal makes more sense, and the latter might make sense in case you have e.g. two stacks with traffic.

jmcs:
Fixes #292 
Fixes #343 

I opted to require the force flag, since several users shown preference for this and it's consistent with deleting multiple stacks. After merging this the delete help of lizzy should also be changed to reflect this change.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -826,7 +826,7 @@ def delete(stack_ref, region, dry_run, force, interactive, ignore_non_existent):
               for stack in cf.get_stacks()
               if matches_any(stack.name, stack_refs)]
 
-    if not (all_with_version(stack_refs) and len(stacks) > 1 and
+    if (not all_with_version(stack_refs) and len(stacks) > 1 and
             not dry_run and not force):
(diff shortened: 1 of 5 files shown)


rafaelcaricio:
👍

PR #352, opened 2016-09-14
Replace scalyr with appdynamics for Spilo

feikesteenbergen:


feikesteenbergen:
👍

senza/templates/postgresapp.py
```diff
@@ -475,7 +471,7 @@ def gather_user_variables(variables, region, account_info):
 
         variables['kms_arn'] = [k['Arn'] for k in kms_keys if k['KeyId'] == kms_keyid][0]
 
-        for key in [k for k in variables if k.startswith('pgpassword_') or k == 'scalyr_account_key']:
+        for key in [k for k in variables if k.startswith('pgpassword_')]:
```
CyberDem0n: What is the point in doing double filtering?
Why not to write something like:
```python
for key, value in variables.items():
    if key.startswith('pgpassword_') and value:
        variables[key] = 'aws:kms:' + encrypt(region=region, KeyId=kms_keyid, Plaintext=value, b64encode=True)
```

feikesteenbergen: https://github.com/zalando-stups/senza/commit/d5bbeaa890ece775eb633afc02fed86ce25f2484

Bugfix for missing scalyr key, so it may be changed again, yes

avaczi:
:+1:

hjacobs:
:+1:

PR #353, opened 2016-09-15
Add pagination to Route53.get_records

### LINKED ISSUE #342: senza traffic displays 0.0% weight for all stacks
When executing `senza traffic [name]` the results displays a weight of 0.0% for all stacks.
`senza domains` shows the weight correctly.

jmcs:
Fixes #342.


jmcs:
See http://boto3.readthedocs.io/en/latest/reference/services/route53.html#Route53.Client.list_resource_record_sets for reference.

rafaelcaricio:
:+1:

PR #358, opened 2016-09-20
Warn users if they are using an old version

### LINKED ISSUE #340: Check if senza is updated
Senza should warn users if there's a new version available in pypi. This check should done only once a day to avoid wasting time and resources.

jmcs:
Fixes #340.


senza/subcommands/root.py
```diff
@@ -1,16 +1,98 @@
...
+    """
+    current_version = LooseVersion(current_version)
+    latest_version = get_latest_version()
+
+    if current_version < latest_version:
+        if __file__.startswith('/home'):
+            # if it's installed in the user folder
+            cmd = "pip install --upgrade stups-senza"
+        else:
+            cmd = "sudo pip install --upgrade stups-senza"
```
rafaelcaricio: I think we should be careful on telling users which command they should use to upgrade Senza. We do not really know how they installed their Senza. Maybe we should simply say that there's a new version available.

rafaelcaricio:
:+1:

senza/subcommands/root.py
```diff
@@ -1,16 +1,98 @@
...
+                latest_version = LooseVersion(str_version)
+    return latest_version
+
+
+def get_latest_version_from_pypi() -> str:
+    """
+    Gets the latest release version pypi api using distutils order
+    to sort the releases (same as pip).
+    """
+    pypi_response = requests.get(PYPI_URL)
```
hjacobs: timeout? error handling?

hjacobs:
I think we need to catch all errors in the version checking --- Senza should never fail or get unnecessary slow just because of PyPI downtimes or similar...

senza/subcommands/root.py
jmcs: Done.

rafaelcaricio:
:+1:

jmcs:
@hjacobs can you re-review this PR. It now [ignores all the exceptions](https://github.com/zalando-stups/senza/pull/358/commits/7ba7b5843fa46aa3ab9151ec6f91fdc71e24debe#diff-cae6888e502721ef34399fe531c1bf28R87).

senza/subcommands/root.py
```diff
@@ -1,16 +1,112 @@
...
 from ..arguments import GLOBAL_OPTIONS, region_option
+from ..error_handling import sentry
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
+PYPI_URL = "https://pypi.python.org/pypi/stups-senza/json"
+ONE_DAY = 86400  # seconds
+
+
+def get_latest_version_from_disc() -> Optional[LooseVersion]:
```
jmcs: Fixed.

hjacobs:
@jmcs let's resolve the conflicts and merge :smile:

jmcs:
@hjacobs also done.

hjacobs:
:+1:

PR #362, opened 2016-09-23
Validate region against a list provided by boto3

jmcs:
Use [`boto3.session.Session.get_available_regions('cloudformation')`](http://boto3.readthedocs.io/en/latest/reference/core/session.html#boto3.session.Session.get_available_regions) to get all valid AWS regions.

### DIFF
--- a/senza/arguments.py
+++ b/senza/arguments.py
@@ -1,18 +1,21 @@
-import re
-
+import boto3.session
 import click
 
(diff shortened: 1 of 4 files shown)


rafaelcaricio:
+1

rafaelcaricio:
:+1:

hjacobs:
:+1:

PR #368, opened 2016-09-29
Use extract_client_error_code everywhere

jmcs:
Reduce code repetition by using extract_client_error_code everywhere senza needs the ClientError code


rafaelcaricio:
👍

PR #372, opened 2016-09-30
#357 lookup security group by tag "Name"

### LINKED ISSUE #357: Lookup SecurityGroups by the tag "Name" rather than GroupName
Both AWS API and CloudFormation allow to refer to a security group by its name if the operation runs in EC2 Classic or the default VPC. Unfortunately it uses the `GroupName` attribute that is automatically generated by AWS if the SG is a part of CloudFormation stack.

It would be a good idea to extend Senza to lookup SG during the CF template generation phase and to use the _tag_ `Name` instead. The tag can be set by another ("system") Senza stack definition, thus allowing the cross-stack references.

Another option would be to support the new cross-stack references that are recently introduced by Amazon:  https://aws.amazon.com/blogs/aws/aws-cloudformation-update-yaml-cross-stack-references-simplified-substitution/

hjacobs:
Fixes #357.

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -55,13 +55,12 @@ def is_status_complete(status: str):
 def get_security_group(region: str, sg_name: str):
     ec2 = boto3.resource('ec2', region)
     try:
-        sec_groups = list(ec2.security_groups.filter(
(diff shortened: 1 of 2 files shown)


rafaelcaricio:
👍

jmcs:
:+1:

PR #374, opened 2016-09-30
Senza version check for TTY only

### LINKED ISSUE #373: Senza version warning should not be printed for non-TTYs (piping result to awk etc)
The Senza version check currently destroys popular shell scripting such as:

```
senza li -o tsv | tail -n +2 | awk '{ print $1 }'
```

We should only print the warning if `sys.stdout` is a TTY.

hjacobs:
Fixes #373 (includes #357).


rafaelcaricio:
👍

jmcs:
:+1:

PR #378, opened 2016-09-30
Improve senza/subcommand/root.py based on pylint

jmcs:
- Added doctstrings
- Used specific variable names
- Formatting
- One sanity check that makes pylint happy

There's still one `exception-too-broad ` warning that was expected because we really want to ignore all exceptions and two errors related to https://github.com/PyCQA/pylint/issues/73


rafaelcaricio:
👍

PR #379, opened 2016-09-30
Improve senza/utils.py based on pylint

jmcs:
- Docstrings;
- Argument name changed in `named_value` (still not really better);
- Removed one step from `camel_case_to_underscore` (@hjacobs I couldn't find any case that failed without it but maybe I'm missing something can you check this change?)

### DIFF
--- a/senza/utils.py
+++ b/senza/utils.py
@@ -19,7 +19,7 @@ def ensure_keys(dict_obj, *keys):
     """
     Ensure ``dict_obj`` has the hierarchy ``{keys[0]: {keys[1]: {...}}}``
 
-    The inner most key will have ``{}`` has value if didn't exist already.
+    The innermost key will have ``{}`` has value if didn't exist already.
     """
     if len(keys) == 0:
         return dict_obj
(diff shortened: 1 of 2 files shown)


senza/utils.py
```diff
@@ -1,12 +1,26 @@
...
+    parameters or Senza Components)
+    """
+    return next(iter(dictionary.items()))
 
 
 def ensure_keys(dict_obj, *keys):
+    """
+    Ensure ``dict_obj`` has the hierarchy ``{keys[0]: {keys[1]: {...}}}``
+
+    The inner most key will have ``{}`` has value if didn't exist already.
```
rafaelcaricio: innermost

rafaelcaricio:
👍

hjacobs:
@jmcs your change will fail for "ThisIsABook" (now it won't generate an underscore between "a" and "book"). So I would say it's better to keep the extra step.

jmcs:
@hjacobs I re-added the extra step and added some unit tests with strings with more than one uppercase in a row.

hjacobs:
:+1:

PR #380, opened 2016-09-30
Improve senza/error_handling.py based on pylint

jmcs:
- Docstrings;
- More specific variable names;
- Reduced some code repetition;

There are still a couple of warnings because `sentry` is constant but lower case, the catch-all except and the typing imports.


rafaelcaricio:
👍

PR #382, opened 2016-10-05
Allow scale down/up specific values for ScaleAdjustment and Cooldown

tmuehl:
Right now in the team we often define the CloudWatch::Alarm and AutoScaling::ScalingPolicy manually in the senza file, not using the senza provided properties, to have different values for scaling up and down. For instance, I often want to scale up much quicker and with more instances than scaling down.

### DIFF
--- a/examples/autoscaling.yaml
+++ b/examples/autoscaling.yaml
@@ -32,11 +32,20 @@ SenzaComponents:
         # default instance count is 1
         # default timeout is 15 minutes
         SuccessRequires: "2 within 5m"
-        # scale two instances up/down at a time
+        # scale two instances up/down at a time (if not overwritten by scale up/down specififc value)
         ScalingAdjustment: 2
+        # scale three instances up at a time (if set, it overwrites the ScalingAdjustment value for scaling up)
+        ScalingUpAdjustment: 3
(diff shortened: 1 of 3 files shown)


examples/autoscaling.yaml
```diff
@@ -32,11 +32,20 @@ SenzaComponents:
         # default instance count is 1
         # default timeout is 15 minutes
         SuccessRequires: "2 within 5m"
-        # scale two instances up/down at a time
+        # scale two instances up/down at a time (if not overwritten by scale up/down specififc value)
         ScalingAdjustment: 2
+        # scale three instances up at a time (if set, it overwrites the ScalingAdjustment value for scaling up)
+        ScalingUpAdjustment: 3
```
tmuehl: I am a bit unhappy abut the name. We now have `ScaleUpThreshold` and `ScaleUpCooldown` but `ScalingUpAdjustment`. I chose that name nevertheless because the existing value is already called `ScalingAdjustment`. Same applies to the scale down pendant.

Any suggestions for more consistent names?

hjacobs:
@tmuehl I renamed it, see #386.

tmuehl:
@hjacobs Thank you 👍

PR #384, opened 2016-10-06
Show helpful error if dns.resolver import fails

jmcs:
I'm singling out this import since it's the only import error I saw on Sentry and there are already multiple reports from multiple hosts.

### DIFF
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -4,8 +4,7 @@ from typing import Dict, Iterator
 
 import boto3
 import click
-import dns.resolver
-from clickclick import Action, action, ok, print_table, warning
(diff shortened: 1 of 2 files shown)


hjacobs:
@jmcs the message should mention the "--force-reinstall" --- the problem is that "dnspython" IS installed but with empty module directory (they changed from dnspython3 to dnspython package IIRC).

jmcs:
@hjacobs thanks. What message do you suggest? "Run (sudo) pip3 install --upgrade ---force-reinstall dnspython"?

rafaelcaricio:
@jmcs please no `sudo`.

hjacobs:
:+1:

PR #386, opened 2016-10-07
Scale up down adjustment

hjacobs:
Basically #382, but with renaming.

### DIFF
--- a/examples/autoscaling.yaml
+++ b/examples/autoscaling.yaml
@@ -37,7 +37,7 @@ SenzaComponents:
         # scale three instances up at a time (if set, it overwrites the ScalingAdjustment value for scaling up)
         ScaleUpAdjustment: 3
         # scale one instance down at a time (if set, it overwrites the ScalingAdjustment value for scaling down)
-        ScaleUpAdjustment: 1
+        ScaleDownAdjustment: 1
         # after scaling, suspend further scaling activities
         # for 30 minutes (1800 seconds)
         # Can be overwritten by scale up/down specific values
--- a/examples/autoscaling.yaml
+++ b/examples/autoscaling.yaml
@@ -32,11 +32,20 @@ SenzaComponents:
         # default instance count is 1
         # default timeout is 15 minutes
         SuccessRequires: "2 within 5m"
-        # scale two instances up/down at a time
+        # scale two instances up/down at a time (if not overwritten by scale up/down specififc value)
         ScalingAdjustment: 2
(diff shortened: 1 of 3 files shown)


examples/autoscaling.yaml
```diff
@@ -32,11 +32,20 @@ SenzaComponents:
         # default instance count is 1
         # default timeout is 15 minutes
         SuccessRequires: "2 within 5m"
-        # scale two instances up/down at a time
+        # scale two instances up/down at a time (if not overwritten by scale up/down specififc value)
         ScalingAdjustment: 2
+        # scale three instances up at a time (if set, it overwrites the ScalingAdjustment value for scaling up)
+        ScaleUpAdjustment: 3
+        # scale one instance down at a time (if set, it overwrites the ScalingAdjustment value for scaling down)
+        ScaleUpAdjustment: 1
```
rafaelcaricio: Do you mean `ScaleDownAdjustment`?

senza/components/auto_scaling_group.py
```diff
@@ -175,27 +175,31 @@ def component_auto_scaling_group(definition, configuration, args, info, force, a
...
         # ScaleUp policy
+        scaling_up_adjustment = int(as_conf.get("ScaleUpAdjustment", default_scaling_adjustment))
         definition["Resources"][asg_name + "ScaleUp"] = {
             "Type": "AWS::AutoScaling::ScalingPolicy",
             "Properties": {
                 "AdjustmentType": "ChangeInCapacity",
-                "ScalingAdjustment": str(scaling_adjustment),
-                "Cooldown": str(as_conf.get("Cooldown", "60")),
+                "ScalingAdjustment": str(scaling_up_adjustment),
+                "Cooldown": str(as_conf.get("ScaleUpCooldown", default_cooldown)),
```
lmineiro: ScaleUpCooldown

lmineiro:
This lgtm but we believe the CF template will actually not use the overwritten Actions. It does apply the alarms, but not the Actions. This is not caused by this PR, just thought that you could check...

tmuehl:
@lmineiro Thanks for pointing out. In this case I have to investigate first.

tmuehl:
@lmineiro Really, is it true?
I believed the name for the `ScalingPolicy` resource is set here: https://github.com/zalando-stups/senza/blob/master/senza/components/auto_scaling_group.py#L180
And here it's used: https://github.com/zalando-stups/senza/blob/master/senza/components/auto_scaling_group.py#L336
Maybe, I missed something.

hjacobs:
:+1:

PR #388, opened 2016-10-10
Use compact JSON format for CF template body in AWS API call

hjacobs:
Use compact JSON format to get more out of the 51,200 bytes CF template limit (see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html).

NOTE: This will not change the pretty-printing of the `print` command.


hjacobs:
:+1:

PR #391, opened 2016-10-11
allow specifying both ELBv2 and ELBv1 for the same ASG (e.g. HTTPS AL…

hjacobs:
This allows using both an ELBv2 ("ALB", application load balancer) and a "classic" ELB with the same ASG.

The ALB only supports HTTP/HTTPS and can therefore not be used in certain situations (e.g. with SSL client certs), i.e. users might want to attach an additional TCP "classic" ELB at the same time.


hjacobs:
:+1:

Issue #387, opened 2016-10-08
Cannot Deploy to Domain Root, No Hosted Zones Found


cjhowald:
We cannot to deploy to our production domain `solutions.zalando.com` due to the following error:

**ERROR: ExternalAppLoadBalancerMainDomain CREATE_FAILED: No hosted zones named zalando.com. found**

It seems senza assumes the domain follows the structure `subdomain`.`hosted_zone` and attempts to create route53 records in `hosted_zone`. This works fine in most cases, but when attempting to deploy to a domain that matches our hosted zone exactly (i.e. the root), it fails because `hosted_zone` is incorrectly set to `zalando.com`. This is consistent with @harti2006's comment on #314.

To confirm this is not just a problem with `solutions.zalando.com`, I tried deploying to the root of our other hosted zone `norris.zalan.do` with the same result.

Is there a way to deploy a stack at our domain root? The announcement of senza 2.0 listed this exact scenario as one of the reasons to move to the new A record setup, so I expect it should be possible.

More details:

- senza version: 2.0.149 
- [senza yaml used](https://github.bus.zalan.do/norris/norris-devops/blob/bugfix/28-change-solution-center-live-url/solution-center/solution-center-prod.yaml).

hjacobs:
@cjhowald this might be a bug, as a workaround you should be able to add another record for the root domain via CF in Senza:

```yaml
Resources:
  RootDomain:
    Type: "AWS::Route53::RecordSet"
    Properties:
      Type: "A"
      Name: "solutions.zalando.com."
      HostedZoneName: "solutions.zalando.com."
      Weight: 0
      SetIdentifier: "solutions-zalando-com-{{SenzaInfo.StackVersion}}"
      AliasTarget:
        HostedZoneId:
          "Fn::GetAtt":
            - "AppLoadBalancer"
            - "CanonicalHostedZoneNameID"
        DNSName:
          "Fn::GetAtt":
            - "AppLoadBalancer"
            - "CanonicalHostedZoneName"

**Comment by cjhowald** [comment]
@hjacobs This workaround appears to be fine for stack creation. But the setup fails when attempting to route traffic.

For example, I created 2 stacks using the above config, but on domain `norris.zalan.do` so as not to mess with our production until this is ironed out. After creation, both stacks show 0 traffic, but really the first stack is getting all the traffic.

```
$ senza traffic user-service-frontend prodtest3
...
[...]

PR #392, opened 2016-10-12
Move AccountArguments to definitions module and document it

jmcs:
Done as part (but does not fix) #387.

This reduces clutter in cli.py.


last-ent:
👍

PR #393, opened 2016-10-12
#370 resolve SSL certs when defining custom ELB Listeners

### LINKED ISSUE #370: Make property overwriting for senza components less destructive
When using a senza component to build a stack, it would be great if it was possible to overwrite properties in a less destructive way. 

### Expected Behavior
We are using the default component `WeightedDnsElasticLoadBalancer`. Now we want to open port 80 in addition to port 443 in order to allow our application to redirect users from http to https. We want to keep all of the configuration and extend it with one additional entry.

### Actual Behavior
Currently this is not possible, because we can only use the option to overwrite the `Listeners` property like this:
```
  - AppLoadBalancer:
      Type: Senza::WeightedDnsElasticLoadBalancer
      ...
      Listeners:
      - LoadBalancerPort: 80
        Protocol: HTTP
        InstancePort: 80
        InstanceProtocol: HTTP
      - LoadBalancerPort: 443
        Protocol: HTTPS
        InstancePort: 80
        InstanceProtocol: HTTP
        SSLCertificateId: "????"
```
Doing this, we completely miss out on the ease of use that the component provides with regard to selecting the correct SSL certificate, because our definition completely overwrites the one senza creates.

--

As a solution it would be nice if we could either flag our `Listeners` entry as "these are additional, not a replacement" or alternatively have a way of tell senza to inject the SSLCertificateId.

hjacobs:
Fixes #370.

### DIFF
--- a/senza/components/elastic_load_balancer.py
+++ b/senza/components/elastic_load_balancer.py
@@ -76,18 +76,26 @@ def get_ssl_cert(subdomain, main_zone, configuration, account_info: AccountArgum
     return ssl_cert
 
 
-def get_listeners(subdomain, main_zone, configuration,
-                  account_info: AccountArguments):
-    ssl_cert = get_ssl_cert(subdomain, main_zone, configuration, account_info)
+def get_listeners(configuration):
     return [
         {
             "PolicyNames": [],
-            "SSLCertificateId": ssl_cert,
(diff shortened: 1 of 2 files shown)


tuxlife:
:+1:

PR #394, opened 2016-10-12
Get hosted zone from route53 when creating template

### LINKED ISSUE #387: Cannot Deploy to Domain Root, No Hosted Zones Found

jmcs:
Fixes #387.

### DIFF
--- a/senza/components/weighted_dns_elastic_load_balancer.py
+++ b/senza/components/weighted_dns_elastic_load_balancer.py
@@ -1,13 +1,26 @@
+from typing import Dict
 
+from senza.definitions import AccountArguments
 from senza.components.elastic_load_balancer import component_elastic_load_balancer
+from senza.manaus.route53 import Route53
 
 
-def component_weighted_dns_elastic_load_balancer(definition, configuration, args, info, force, account_info,
+def component_weighted_dns_elastic_load_balancer(definition,
+                                                 configuration: Dict,
+                                                 args,
+                                                 info,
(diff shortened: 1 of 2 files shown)


jmcs:
I already tested senza print and create with this change and I got the same results for the stacks I've but it should also support stacks that use a domain with the same name as the hosted zone.

ainmosni:
:+1:

PR #400, opened 2016-10-20
add SubnetAutoConfiguration component with PublicOnly option

hjacobs:
For our Kubernetes setup: allow configuring public subnets without Taupage


hjacobs:
:+1:

PR #406, opened 2016-10-27
use dnspython instead of deprecated dnspython3

hjacobs:
dnspython3 has been superseded by the regular dnspython kit, which now supports both Python 2 and Python 3.


hjacobs:
:+1:

PR #411, opened 2016-11-01
bugfix: added posibility to set vpc id for elbv2

adyach:


senza/components/elastic_load_balancer_v2.py
```diff
@@ -96,6 +96,8 @@ def component_elastic_load_balancer_v2(definition,
     else:
         loadbalancer_subnet_map = "LoadBalancerSubnets"
 
+    vpc_id = configuration.get("VpcId") or account_info.VpcID
```
hjacobs: actually `configuration.get("VpcId") or account_info.VpcID` is better as this will also work if VpcId is  false/empty

hjacobs:
@adyach can you add a test please?

adyach:
@hjacobs 5a73fb8

hjacobs:
:+1:

PR #414, opened 2016-11-03
#402 find latest AMI for CoreOS

### LINKED ISSUE #402: CoreOS Auto Configuration Component
We are currently testing [Kubernetes on AWS](https://github.com/zalando-incubator/kubernetes-on-aws) with CoreOS AMIs. We take the AMI ID from https://coreos.com/dist/aws/aws-stable.json and put it into the Senza definition YAML manually. It would be better to have a Senza component to automatically find and use the latest CoreOS AMI (similar to what the `StupsAutoConfiguration` component does for Taupage).

hjacobs:
Fixes #402


hjacobs:
:+1:

PR #420, opened 2016-11-25
Wait stacks to be ready before changing traffic

### LINKED ISSUE #398: Senza traffic should wait for stack to be ready to change traffic
### Current Behaviour

When running change traffic `senza traffic` on stack recently deployed. The status code of the stack is not ready (`(CRETE|UPDATE)_COMPLETE`) and Senza fails trying to update it.
### Cause of the Problem

It is not possible to update CF stack when it's status code is not in `(CRETE|UPDATE)_COMPLETE` (or we should not change when it is not in one of those status code).
### Expected Behaviour

Senza should wait the stack to get in a final status code if the current status of the stack is `(CREATE|UPDATE)_IN_PROGRESS` and ending up in some good state `(CREATE|UPDATE)_COMPLETE`.

rafaelcaricio:
Fixes #398

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1172,7 +1172,7 @@ def traffic(stack_name, stack_version, percentage, region, output, interval):
                                 # some operation in progress, let's wait some time to try again
                                 all_stacks_in_final_state = False
                                 info(
-                                    "Waiting stack {} ({}) to perform traffic change...".format(
+                                    "Waiting for stack {} ({}) to perform traffic change..".format(
                                         related_stack.StackName, current_stack_status))
                                 time.sleep(interval)
                     else:
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1141,9 +1141,10 @@ def domains(stack_ref, region, output, w, watch):
 @output_option
 @stacktrace_visible_option
 def traffic(stack_name, stack_version, percentage, region, output):
-    '''Route traffic to a specific stack (weighted DNS record)'''
+    """Route traffic to a specific stack (weighted DNS record)"""
 
     stack_refs = get_stack_refs([stack_name, stack_version])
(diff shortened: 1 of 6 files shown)


senza/cli.py
```diff
@@ -1152,29 +1153,32 @@ def traffic(stack_name, stack_version, percentage, region, output):
...
+                    if len(related_stacks) > 0:
+                        for related_stack in related_stacks:
+                            current_stack_status = related_stack.StackStatus
+
+                            if current_stack_status.endswith('_COMPLETE') or current_stack_status.endswith('_FAILED'):
+                                continue
+                            elif current_stack_status.endswith('_IN_PROGRESS'):
+                                # some operation in progress, let's wait some time to try again
+                                all_stacks_in_final_state = False
+                                info(
```
lukasniemeier-zalando: This can get a bit spammy right? :wink:

lukasniemeier-zalando:
:+1:

rafaelcaricio:
👍

senza/cli.py
rafaelcaricio: @lukasniemeier-zalando now I have added an option to be able to set the wait interval. So you control the spammy behaviour. ;)

senza/cli.py
```diff
@@ -1152,29 +1156,32 @@ def traffic(stack_name, stack_version, percentage, region, output):
...
+                                    "Waiting stack {} ({}) to perform traffic change...".format(
```
hjacobs: The grammar sounds wrong: better: "Waiting for stack ... to perform traffic change.." (add "for" and use only two trailing dots as we do everywhere)

hjacobs:
@rafaelcaricio you need to fix tests too..

hjacobs:
:+1:

PR #434, opened 2017-01-19
Add param --update-if-exists to senza create to make it idempotent

### LINKED ISSUE #416: add param --update-if-exists to senza create to make it idempotent
It would be nice if we got the same result when calling `senza create` several times with the same parameters.
At the moment, the second time `senza create` is called, we get an error 'Stack already exists`.

It would be useful to have a parameter like `--update-if-exists`.

That might require to add a function to check if the stack name exists, and redirect to `senza update` if it returns True. 

Right now we have a bit of bash to do this:
```
stack_name=$(grep 'StackName:' "${SENZA_TEMPLATE}" | awk '{print $2}')

# check if stack exists
if [[ ! -z $(senza --region ${EC2_REGION} list --output tsv | awk '{print $1 ":" $2}' | grep ${stack_name}:${STACK_VERSION}) ]]; then
  echo "Stack '${stack_name}' exits. Updating..."
  senza_command=update
else
  echo "Stack '${stack_name}' does not exit. Creating..."
  senza_command=create
fi

# create/update stack
senza "${senza_command}" \
    --stacktrace-visible \
    --region "${REGION}" \
    --parameter-file "${SENZA_PARAM_FILE}" \
    "${SENZA_TEMPLATE}" \
    "${STACK_VERSION}" \
    dockerTag="${TAG}"
```

However in a Makefile, this gets so ugly that we put it in a separate shell script. 
And that results in having to pass all those variables (REGION, SENZA_PARAM_FILE, SENZA_TEMPLATE, STACK_VERSION, TAG) from the Makefile to the shell script, which means the script itself becomes something to maintain... 

It'd be nice to just have this:
```
senza create \
    --update-if-exists \
    --stacktrace-visible \
    --region "${REGION}" \
    --parameter-file "${SENZA_PARAM_FILE}" \
    "${SENZA_TEMPLATE}" \
    "${STACK_VERSION}" \
    dockerTag="${TAG}"
```

jmcs:
Fixes #416.

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -6,12 +6,14 @@ import base64
 import collections
 import functools
 import re
+from pprint import pformat
 
 import arrow
 import boto3
 import yaml
 from botocore.exceptions import ClientError, BotoCoreError
 from click import FileError
+from clickclick import Action, info
 
 from .exceptions import SecurityGroupNotFound
 from .manaus.boto_proxy import BotoClientProxy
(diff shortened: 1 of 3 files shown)


rafaelcaricio:
👍

Issue #436, opened 2017-01-24
Senza still suggests 'mai' tool


outime:
When AWS credentials are expired:

`Use the "mai" command line tool to get a new temporary access key.`

Need to be updated to suggest `zaws`.

outime:
@jmcs we made a PR at the same time, I just closed mine so yours is the only one open now.

PR #437, opened 2017-01-24
#436 Updated suggestion from 'mai' to 'zaws'

outime:


outime:
Just realized there's exactly one PR for this already :D

PR #438, opened 2017-01-24
Suggest zaws instead of mai

### LINKED ISSUE #436: Senza still suggests 'mai' tool

jmcs:
Fixes #436


last-ent:
👍

PR #440, opened 2017-01-26
Support for additional properties, eg. IdleTimeout of load balancer

grzegorzjeziorski:
I added one additional line for supporting http://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes in ELBv2


hjacobs:
Did you test with the empty `{}`?

grzegorzjeziorski:
Should be fixed now. I tested both variants, with and without a property set.

hjacobs:
:+1:

jmcs:
:+1:

PR #442, opened 2017-01-30
do not fail if no updates are to be performed

hjacobs:
This copies the logic from manaus/cloudformation: https://github.com/zalando-stups/senza/blob/master/senza/manaus/cloudformation.py#L199


hjacobs:
:+1:

Issue #441, opened 2017-01-27
Automatically login to Pier One


musiKk:
When creating a stack, Pier One is checked. This is sensible. If I'm not logged in, senza barfs and suggests to log in. Senza should try to log the user in first and only if that fails abort with an error.

I think the behavior of the `--force` flag should stay the same, i.e., when it is used, don't even try to check for the existence of the image and thus don't try to log the user in.

akauppi:
Personally, I like to keep log ins away from any scripts, and would like for the current separation to remain.

The reason is that I would like to be personally aware of when such logins happen, and if any or all the tools start logging in on my behalf, this control essentially vanishes.

musiKk:
It seems there is already a proposed change.

Personally I don't really see a problem though. I mean, the tools are interwoven to a degree and the only thing I'm doing after I get the error is to do what it suggests (login to Pier One) and retry stack creation. When would attempting to log in not be a sensible solution? I think this is a matter of purity (separate all the things) vs. usability (do what I want!) and I'm in favor of usability.

musiKk:
I think this is deployed. I'll just clean up after myself.

PR #444, opened 2017-02-01
#441 Try to log into Pier One if the existing token has expired.

last-ent:
Instead of throwing an error if the existing token expires, now we try to re-login and return the token.


rafaelcaricio:
👍

Issue #443, opened 2017-02-01
Feature Request: Ability to place a stack into maintenance mode


imduffy15:
I have many alarms for my services within ZMON, when deploying upgrades I would like to not trip those alarms.

I would like to be able to deploy the new services with maintenance mode enabled(just an autoscaling group tag that gets populated to the instances) so the ZMON check will exclude the entity.

When the service is deployed and ready to handle traffic I would like to disable maintenance mode(Change the tag to false) for the autoscaling group and all associated instances.

Thoughts?

musiKk:
In order to be consistent with other commands like traffic or scale, `senza maintenance foo` should tell what the current state is and something like true/false, 1/0, on/off, ... should be used to update.

PR #445, opened 2017-02-01
Ability to place a stack into maintenance mode #443

imduffy15:
Given a stack it will detect all asgs within that stack negates the tag named 'maintenance' and do the same on all instances currently belonging to that asg.

$ senza maintenance some-stack-name
Changing maintenance mode on Auto Scaling Group some-stack-name-version-AppServer-1R2KFL5HLI5NT..
Maintance mode set to False for some-stack-name-version-AppServer-1R2KFL5HLI5NT

$ senza maintenance some-stack-name
Changing maintenance mode on Auto Scaling Group some-stack-name-version-AppServer-1R2KFL5HLI5NT..
Maintance mode set to True for some-stack-name-version-AppServer-1R2KFL5HLI5NT

Signed-off-by: Ian Duffy <ian.duffy@zalando.ie>

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import ast
 import base64
 import calendar
 import collections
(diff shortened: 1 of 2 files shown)


hjacobs:
Hmm, I'm not sure if this feature makes sense to have in Senza directly. It sounds very special and application specific (some app needs to actually interpret the AWS tag).

It also breaks Cloud Formation update handling as tags are directly manipulated via the AWS API.

imduffy15:
> Hmm, I'm not sure if this feature makes sense to have in Senza directly

Thought this might be the case alright. Didn't see any better place to put it other than a team specific script but that required a bit of re-work to get the base thats already in senza.

> it sounds very special and application specific

The idea would be to have the ZMON checks looking for the maintenance tag.

> It also breaks Cloud Formation update handling as tags are directly manipulated via the AWS API.

As does patch :)

hjacobs:
@imduffy15 I'm not totally opposed, but I would at least like to see others supporting (and using) your idea. It probably makes no sense to have this code in Senza for only one user.

The command is not idempotent as it toggles the flag, this might be dangerous when used in scripts (?).

imduffy15:
> The command is not idempotent as it toggles the flag, this might be dangerous when used in scripts (?).

Okay, I'll add a parameter for on/off.

imduffy15:
Discussed in zmon chat with @Jan-M  he suggested using https://docs.zmon.io/en/latest/user/downtimes.html however that will not work for entities that do not yet exist.

hjacobs:
I'm closing this as it got no support from other users and I don't want to merge something specifically for one (or a few) users only.

PR #447, opened 2017-02-09
Only waits for traffic change if timeout provided

rafaelcaricio:
Related to https://github.com/zalando/lizzy/issues/194

Only waits stacks to be ready if specified the timeout period. By default behaves the same way as before, trying to exectute the traffic change immediately.

### DIFF
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -469,10 +469,7 @@ def all_stacks_in_final_state(related_stacks_refs: list, region: str, timeout: O
 
             for related_stack in related_stacks:
                 current_stack_status = related_stack.StackStatus
-
-                if current_stack_status.endswith('_COMPLETE') or current_stack_status.endswith('_FAILED'):
-                    continue
-                elif current_stack_status.endswith('_IN_PROGRESS'):
+                if current_stack_status.endswith('_IN_PROGRESS'):
                     # some operation in progress, let's wait some time to try again
                     all_in_final_state = False
                     info(
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -463,7 +463,7 @@ def all_stacks_in_final_state(related_stacks_refs: list, region: str, timeout: O
             all_in_final_state = True
             related_stacks = list(get_stacks(related_stacks_refs, region))
 
-            if len(related_stacks) <= 0:
+            if not related_stacks:
                 error("Stack not found!")
                 exit(1)
 
--- a/senza/aws.py
+++ b/senza/aws.py
@@ -7,13 +7,17 @@ import collections
 import functools
 import re
 from pprint import pformat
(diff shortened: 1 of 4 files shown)


senza/cli.py
```diff
@@ -1144,54 +1144,80 @@ def domains(stack_ref, region, output, w, watch):
                         rows, styles=STYLES, titles=TITLES)
 
 
+@contextmanager
+def all_stacks_in_final_state(stack_name: str, region: str, timeout: Optional[int], interval: int):
```
jmcs: this shouldn't be in cli.py

rafaelcaricio: Would you mind to suggest a place where you think it should go?

jmcs:
:+1:

senza/aws.py
```diff
@@ -436,3 +440,48 @@ def update_stack_from_template(region: str, template: dict, dry_run: bool):
...
+                    if current_stack_status.endswith('_COMPLETE') or current_stack_status.endswith('_FAILED'):
+                        continue
+                    elif current_stack_status.endswith('_IN_PROGRESS'):
+                        # some operation in progress, let's wait some time to try again
+                        all_in_final_state = False
+                        info(
+                            "Waiting for stack {} ({}) to perform requested operation..".format(
+                                related_stack.StackName, current_stack_status))
+                        time.sleep(interval)
+            else:
```
last-ent: Have you considered writing code from 466-480 as follows?

```
if len(related_stacks) == 0:
    error('Stack not found!')
    exit(1)
for related_stack in related_stacks:
    #...
```

It reduces the amount of indentation in code which makes it more readable IMO.

rafaelcaricio: Good catch.

senza/aws.py
```diff
@@ -436,3 +440,48 @@ def update_stack_from_template(region: str, template: dict, dry_run: bool):
...
+            if len(related_stacks) <= 0:
+                error("Stack not found!")
+                exit(1)
+
+            for related_stack in related_stacks:
+                current_stack_status = related_stack.StackStatus
+
+                if current_stack_status.endswith('_COMPLETE') or current_stack_status.endswith('_FAILED'):
+                    continue
+                elif current_stack_status.endswith('_IN_PROGRESS'):
```
last-ent: Should there be an `else` clause here? Because if only three statuses are possible, then a simple `if ...endswith('_IN_PROGRESS')` would suffice, right?

rafaelcaricio: The simple if should surface, indeed. I will make this change.

last-ent:
👍

jmcs:
:+1:

jkozlowicz:
👍

Issue #451, opened 2017-03-02
Exporting ApplicationName as an AWS tag for ELBs


jkozlowicz:
We need a way to identify ELBs of multiple instances of the same application (i.e. multiple instances of `LizzyAgent` hosted in different AWS accounts).

I added a way to do this by setting `ApplicationName` attribute in `SenzaInfo` section:
```
# basic information for generating and executing this definition
SenzaInfo:
  StackName: hello-<team_name>
  ApplicationName: hello-app
```
which will be exported as an AWS tag for the ELB.

PR #452, opened 2017-03-02
Exporting ApplicationName as an AWS tag for ELBs

jkozlowicz:
I added a way to export `ApplicationName` attribute in `SenzaInfo` section as an AWS tag for the ELB.

More info: #451

### DIFF
--- a/senza/components/elastic_load_balancer.py
+++ b/senza/components/elastic_load_balancer.py
@@ -4,6 +4,8 @@ from senza.aws import resolve_security_groups
 from senza.definitions import AccountArguments
 
 from ..cli import TemplateArguments
+from ..utils import extract_attribute
+
 from ..manaus import ClientError
 from ..manaus.acm import ACM, ACMCertificate
 from ..manaus.iam import IAM, IAMServerCertificate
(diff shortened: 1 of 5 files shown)


hjacobs:
That code looks tailored to a very specific use case. Why was `StackName` not good enough?

hjacobs:
Or why not add Tags in your Senza def directly (not requiring Senza code changes)?

jkozlowicz:
`StackName` was not enough because it can vary across AWS accounts of different teams - `LizzyAgent` has its `StackName` set to `lizzy-<team_name>`.

We need a unified way to target all ELBs of `LizzyAgents` in `ZMON`. This way we can target all entities with `application_name` equal to `X`.

I was not aware that tags could be added in senza definition in a way that then they can be used to filter entities in `ZMON`.

hjacobs:
I think the PR makes no sense as it is now (too specific). We should try to add the option to add Tags to the ALB (if this does not work yet?).

jkozlowicz:
@hjacobs yes you are right. We were able to add tags to ALB in senza definition for `Lizzy` so this one can be closed.

hjacobs:
@jkozlowicz cool, thanks.

PR #455, opened 2017-03-16
Don't try to change traffic unless really needed

### LINKED ISSUE #453: Traffic switch affected by older stacks being deleted in background
Just had a new version of the test stack deployed via CI/CD.  In the CD deploy step the traffic is switched automatically 100% to the newest version and the oldest one is scheduled for deletion.

Noticed a problem in the newest version, so wanted to switch traffic back to old stable one.  Observed the current traffic report:
```
$ senza traffic abcdef-test
Stack Name │Version│Identifier       │Weight%
abcdef-test cd209   abcdef-test-cd209     0.0 
abcdef-test cd210   abcdef-test-cd210     0.0 
abcdef-test cd212   abcdef-test-cd212   100.0 
```

Issued command to move 100% of the traffic from cd212 back to cd210.  The oldest one, cd209 should not be involved in any way, however:
```
$ senza traffic abcdef-test cd210 100
Calculating new weights.. OK
Stack Name │Version│Identifier       │Old Weight%│Delta │Compensation│New Weight%│Current
abcdef-test cd209   abcdef-test-cd209         0.0                             0.0         
abcdef-test cd210   abcdef-test-cd210         0.0  100.0                    100.0 <       
abcdef-test cd212   abcdef-test-cd212       100.0 -100.0                      0.0         
Setting weights for abcdef-test.acid.zalan.do...Validation Error: Stack:arn:aws:cloudformation:eu-central-1:1234567890:stack/abcdef-test-cd209/8ca341d0-0278-11e7-9750-500c52a6ce9a is in DELETE_IN_PROGRESS state and can not be updated.
```

The update of cd210 and cd212 still happened, but the error here seems to be avoidable.  We know that we need to move the traffic only between the two stacks, cd209 should not be touched at all.

jmcs:
Fixes #453 

This pull request also removes two unused parameters in ``traffic.set_new_weights``

### DIFF
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -134,8 +134,10 @@ def compensate(calculation_error, compensations, identifier, new_record_weights,
     return percentage
 
 
-def set_new_weights(dns_names: list, identifier, lb_dns_name: str,
-                    new_record_weights: Dict, region: str):
+def set_new_weights(dns_names: list,
+                    old_record_weights: Dict,
+                    new_record_weights: Dict,
+                    region: str):
     action('Setting weights for {dns_names}..', dns_names=', '.join(dns_names))
     for idx, dns_name in enumerate(dns_names):
         domain = dns_name.split('.', 1)[1]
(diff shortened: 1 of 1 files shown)


rafaelcaricio:
👍

jkozlowicz:
👍

PR #458, opened 2017-03-22
Refactored get_load_balancer_name to be used when getting stack health

arjunrn:
Currently when getting stack health through `senza health` the LB metrics cannot be fetched for stacks with long application names. This is because the application name is truncated when creating the load balancer. However in the `health` command the name truncation is not performed which means the metrics are not fetched correctly. This PR moves the truncation function to the `utils` package so that it can be used in the `cli` module as well as the `elb` module.


arjunrn:
@rafaelcaricio @jmcs @hjacobs Could I get some feedback on this. We want to use this in our deployment pipelines and we are waiting for the feature to be merge.

hjacobs:
Thanks for the PR! :smile:

hjacobs:
Released in https://github.com/zalando-stups/senza/releases/tag/2.1.71

PR #460, opened 2017-04-06
Adding the 'TargetGroup' for ALB inside the loop

duergner:
When using an `ElasticLoadBalancerV2` the `TargetGroup` name was not appended on the actual ref but rather tried to be added to the list itself which causes a Python error to be thrown.

Looked like a copy and past error from the `ElasticLoadBalancer` version.


hjacobs:
I'm wondering how this could have worked before?

hjacobs:
Can you add a test for this case?

duergner:
Sure - will push tomorrow morning.

The syntax for the old ELB version is slightly different as there's been no need to append the `TargetGroup` suffix. I assume it never worked for ALBs

duergner:
I added test cases for both `ElasticLoadBalancer` and `ElasticLoadBalancerV2` references for the three different cases each I believe should provide full coverage

hjacobs:
:+1:

hjacobs:
I'm fixing the tests (rely on date :disappointed: )

hjacobs:
@duergner I fixed the tests in #462, please rebase your PR..

duergner:
@hjacobs done

hjacobs:
:+1:

hjacobs:
Released: https://github.com/zalando-stups/senza/releases/tag/2.1.70

PR #463, opened 2017-04-18
Proper error if parameter file is empty

jmcs:
If a parameter file is empty senza will currently raise the following exception:

```
AttributeError: 'NoneType' object has no attribute 'items'
```


tuxlife:
:+1:

PR #464, opened 2017-05-11
Allow filtering of fields returned by senza list

inikolaev:
Right now `senza list` returns a fixed list of fields for each stack. There're use cases where I might only need some of them, e.g. `version`.

I'm proposing to add `--field` option which would allow to specify which field to include in the output:

```bash
$ senza list -o tsv my-stack --field version
version
1
```

It's possible to specify parameter multiple times in the command line if more parameters needed:

```bash
$ senza list -o tsv my-stack --field version --field stack_name
stack_name      version
my-stack        1
```

If no `--field` parameter specified, the tool will output all fields as before:

```bash
$ senza list -o tsv my-stack
stack_name      version status  creation_time   description
my-stack        1       UPDATE_COMPLETE 7h ago  My Stack
```

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -403,9 +403,10 @@ def all_with_version(stack_refs: list):
 @watch_option
 @watchrefresh_option
 @click.option('--all', is_flag=True, help='Show all stacks, including deleted ones')
+@click.option('--field', '-f', metavar='NAME', multiple=True, help='Specify field to be returned')
 @click.argument('stack_ref', nargs=-1)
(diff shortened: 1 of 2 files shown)


hjacobs:
Nice! :smile: Does it make sense to add this to other commands as well (e.g. `instances`)?

hjacobs:
:+1:

inikolaev:
I guess it does make sense, I just don't know all commands and mostly use `senza list`, but can try to take a look.

PR #465, opened 2017-05-12
Filter output fields

inikolaev:
Added column filtering to all commands which have that functionality

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -116,6 +116,13 @@ MAX_COLUMN_WIDTHS = {
 }
 
 
+def filter_output_columns(output_columns, filter_columns=list()):
+    if filter_columns:
+        return [column for column in output_columns if column in filter_columns]
+
+    return output_columns
+
+
 def print_json(data, output=None):
     if output == 'yaml':
         parsed_data = yaml.safe_load(data)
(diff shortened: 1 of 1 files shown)


senza/cli.py
```diff
@@ -116,6 +116,13 @@
 }
 
 
+def filter_output_columns(output_columns, filter_columns=list()):
```
hjacobs: please use the singleton `None` as the default parameter, never an actual instance (does not make a difference here but is important in general as the default value is only instantiated once!)

senza/cli.py
```diff
@@ -510,12 +513,13 @@ def get_classic_load_balancer_metrics(cloudwatch, lb_name, start, now):
 @cli.command()
 @region_option
 @output_option
+@click.option('--field', '-f', metavar='NAME', multiple=True, help='Specify field to be returned')
```
hjacobs: you can define the option once at the top of the file and reuse it (like `@region_option`)

hjacobs:
Thanks for the PR! Please see my comments.

senza/cli.py
inikolaev: True, I don't need a default value at all as I always expect this function to be called with 2 arguments.

senza/cli.py
inikolaev: Nice!

hjacobs:
Thanks for the PR, will review later..

inikolaev:
@hjacobs do you have any more comments?

hjacobs:
:+1:

PR #466, opened 2017-06-06
Changing flow for setting traffic

duergner:
This will collect all stack updates and only execute them on batch in the end to avoid having problems due to `UPDATE_IN_PROGESS` when setting traffic on a stack that has more than one ELB/ALB attached.


duergner:
It's the current WiP - I'm open for any feedback on the overall approach and ideas.

/cc @hjacobs

senza/traffic.py
```diff
@@ -183,28 +184,42 @@ def set_new_weights(dns_names: list,
...
-                    if r.set_identifier == stack_name:
-                        record = r
-                        break
-                if record and record.weight != percentage:
-                    record.weight = percentage
-                    hosted_zone.upsert([record],
-                                       comment="Change weight of {} to {}".format(stack_name,
-                                                                                  percentage))
-                    changed = True
+            if not stack_name in updates.keys():
```
hjacobs: `if stack_name not in updates`

hjacobs:
Some test is failing..

duergner:
Going to have a look at the failing test today.

duergner:
@hjacobs I did some digging into the failing test (the one that did not fail previously but fails now)

```
    # test not changed
    for stack in m_stacks.values():
        stack.update.side_effect = StackNotUpdated('abc')

    m_ok = MagicMock()
    monkeypatch.setattr('senza.traffic.ok', m_ok)

    with runner.isolated_filesystem():
        run(['v4', '100'])
        m_ok.assert_called_once_with(' not changed')
```

I think that test case is not really testing the desired behaviour  as far as I understand and see the flow. Traffic on `v4` should not be `100` before, i.e. the command should actually update something. Though I don't have any idea yet why the test is passing on for `master`.

duergner:
We had some chat today about that failing test in the office and I think we now understand what is causing the changed behaviour.

As we set the side effect explicitly on the mock the actual stack update was been *executed* anyway with the old code which which resulted in the `except` clause not *executing* the manual Route53 action which would have triggered the changed flag to be set.

As the new code flow puts the check whether or not a Route53 update need to happen prior to executing the stack update, it is scheduled to be executed whenever the *StackNotUpdated* exception is encountered regardless of the state of the weights at this point.

Basically the test is testing / assuming a behaviour which would never happen in real life (to my understanding) this failing test is *correct* now.

Your thoughts on this @hjacobs

tests/test_cli.py
```diff
@@ -1406,7 +1406,12 @@ def update_weights(stacks):
 
     with runner.isolated_filesystem():
         run(['v4', '100'])
-        m_ok.assert_called_once_with(' not changed')
+        # TODO this now only checks for the correct result but it should actually check for the raw Route53 commands
+        assert get_weight(m_stacks['myapp-v1']) == 0
+        assert get_weight(m_stacks['myapp-v2']) == 0
+        assert get_weight(m_stacks['myapp-v3']) == 0
+        assert get_weight(m_stacks['myapp-v4']) == 200
+        # m_ok.assert_called_once_with(' not changed')
```
tuxlife: Please remove dead code

tuxlife:
:+1:

tuxlife:
Please use Squash-Merge

duergner:
@tuxlife will one of you maintainer do a squash merge or shall I make a new PR with a squashed branch?

PR #469, opened 2017-06-15
remove dns.resolver import error handling

hjacobs:
The import error handling is no longer needed (the use case was the migration from dnspython3 to dnspython).

This fixes the failing unit test.


hjacobs:
:+1:

PR #472, opened 2017-06-20
Fix issue with changed set traffic handling

duergner:
The PR which introduced the changed handling of set traffic solving the UPDATE_IN_PROGRESS issue with stacks having multiple ELBs (and therefore multiple DNS names) did treat a `stack` object returned by `CloudFormationStack.get_by_stack_name` as if it would be the same object being returned by subsequent calls with the same stack_name. As this is not the case this leads to only the first update being actually executed.

### DIFF
--- a/senza/traffic.py
+++ b/senza/traffic.py
@@ -146,8 +146,11 @@ def set_new_weights(dns_names: list,
                 # Stack weight will not change
                 continue
             try:
-                stack = CloudFormationStack.get_by_stack_name(stack_name,
-                                                              region=region)
+                if stack_name not in updates.keys():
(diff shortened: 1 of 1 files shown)


jmcs:
Can you include some tests for this?

duergner:
I can do so yes - any chance to merge that PR and have the tests on a follow up PR?

jmcs:
:+1:

duergner:
@jmcs @hjacobs any chance this PR could get merged and released?

jmcs:
:+1:

last-ent:
👍

PR #477, opened 2017-07-12
Put nat_gateways into UserData if instance runs without public ip in DMZ

CyberDem0n:
We have a use-case when we want to run instances in DMZ or public subnet but without public ips. Elastic ip would be assigned later to one of the instances.
Without public ip instance will not be able to initialize (download docker, push logs, use AWS api, etc...). To solve this problem taupage will create separate routing table for outgoing https requests and use a custom nat gateway for that.

### DIFF
--- a/senza/components/taupage_auto_scaling_group.py
+++ b/senza/components/taupage_auto_scaling_group.py
@@ -1,3 +1,4 @@
+import boto3
 import json
 import re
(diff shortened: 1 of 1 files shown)


jmcs:
:+1:

CyberDem0n:
:+1:

Issue #479, opened 2017-07-19
if default param is set, empty param can't overwrite it


drummerwolli:
if i set in my deployment yaml file a default value for a parameter, i need to specify a new parameter with `>= len() = 1` to overwrite the default value. It's not possible to overwrite the default value with an empty parameter.

Example:

Imagine this head of a senza deployment file (reduced version for simplicity):

```
SenzaInfo:
  StackName: hello-world
  Parameters:
    - Parameter1:
        Description: "Test Parameter"
        Default: "foobar"
```
if i execute senza create (or print) and provide no parameter, then the default param is used (which is correct). also this command executes as expected:

`senza create test.yaml 1 Parameter1="barfoo"`

BUT if i don't want that parameter to be set (it should be empty or None), then i would expect this to execute:

`senza create test.yaml 1 Parameter1=""`

But then senza is taking the default value ...

Usecase: In our application we have a feature toggle based on those Parameters. As soon as something is set here (in this case a password), the feature is activated. But if you want to deactivate the feature for a specfic deployment, you want to "unset" this parameter. for documentation reasons though, we have an example in our deployment file in the `Default` section (it's kind of a convention to use this field for documentation purposes as well, right?), so we can't unset this right now ...

Is this intended? if not i could take a look, if someone could point me to the location of this specific evaluation. I only did a first rough check, but could not find a `if param:` clause somewhere (which i think is causing this) ...

jmcs:
Fixed by #486

PR #485, opened 2017-08-02
consider params can be empty

drummerwolli:
this solves #479 ...


drummerwolli:
i think the failing test is in general failing (also looking at https://github.com/zalando-stups/senza/pull/484) ...

hjacobs:
See #486 with master branch merged (fixed tests).

Issue #478, opened 2017-07-12
Senza scale should ask when scaling multiple versions


elezar:
Yesterday, I was scaling a stack with two versions and executed:
```
$ senza scale STACKNAME 0
```
this scaled **BOTH** versions of the stack to 0 instances. I corrected this with a subsequent scale command, but this left the stack in a state where the minimum number of instances was 0 and this caused an incident later at night.

**My expectation** was that `scale` would behave the same as `delete` in that it asks for confirmation when scaling a stack with multiple versions present.

vadeg:
This issue was merged and can be closed, no?

elezar:
Thanks. Closing this issue.

(Note I don't have a means to currently test this, but I did have a look at the merged PR)

PR #488, opened 2017-09-21
#478 - 'senza scale' prompt when more than one instance might be scaled.

vadeg:
#478 fixed. 

* User should confirm operation when more than one stack found to scale. 
* `--force` flag used to turn off confirmation and scale all found instances.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1473,9 +1473,10 @@ def respawn_instances(stack_ref, inplace, force, region):
 @cli.command()
 @click.argument('stack_ref', nargs=-1)
 @click.argument('desired_capacity', type=click.IntRange(0, 100, clamp=True))
+@click.option('--force', ' /-F', default=False, help='Force scaling multiple stacks if needed')
 @region_option
 @stacktrace_visible_option
-def scale(stack_ref, region, desired_capacity):
+def scale(stack_ref, region, desired_capacity, force):
     '''Scale Auto Scaling Group to desired capacity'''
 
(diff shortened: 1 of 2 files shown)


senza/cli.py
```diff
@@ -1484,6 +1485,11 @@ def scale(stack_ref, region, desired_capacity):
 
     asg = BotoClientProxy('autoscaling', region)
 
+    stack_count = len(list(get_stacks(stack_refs, region)))
+    if not force and stack_count > 1:
+        confirm_str = 'Number of stacks to be scaled - %s. Do you want to continue?' % stack_count
```
jmcs: use [.format](https://docs.python.org/3/library/string.html#string-formatting) instead of the legacy `%` formatting.

senza/cli.py
```diff
@@ -1473,9 +1473,10 @@ def respawn_instances(stack_ref, inplace, force, region):
 @cli.command()
 @click.argument('stack_ref', nargs=-1)
 @click.argument('desired_capacity', type=click.IntRange(0, 100, clamp=True))
+@click.option('--force', ' /-F', default=False, help='Force scaling multiple stacks if needed')
```
jmcs: Please define ``--force`` as flag like [other subcommands](https://github.com/vadeg/senza/blob/128ac6540138cadfbfc8c112a7dbb8c078adced3/senza/cli.py#L1455).

jmcs:
:+1:

lotharschulz:
👍

hjacobs:
:+1:

Issue #493, opened 2017-11-28
Allow to specify Redis cluster name when using RedisNode component


inikolaev:
Right now when using `RedisNode` component to create Redis instance cluster name is set to the name of the resource section where component is declared:

```yaml
SenzaComponents:
  - Configuration:
      Type: Senza::StupsAutoConfiguration
  - Redis:
      Type: Senza::RedisNode
      CacheNodeType: "{{Arguments.InstanceType}}"
      CacheParameterGroupName: "{{Arguments.RedisParameterGroupName}}"
      EngineVersion: "{{Arguments.RedisVersion}}"
      SecurityGroups:
        - Fn::GetAtt: ["RedisSecurityGroup", GroupId]
```

The above definition will create a cluster named `redis`.

While it's possible to use mustache templates to name a section I don't particulary like that approach as it hurts readability in my opinion.

I would like to propose to support `ClusterName` attribute which would allow to set cluster name, For compatibility it will fallback to using section name. After that change the template will look like this:

```yaml
SenzaComponents:
  - Configuration:
      Type: Senza::StupsAutoConfiguration
  - Redis:
      Type: Senza::RedisNode
      CacheNodeType: "{{Arguments.InstanceType}}"
      CacheParameterGroupName: "{{Arguments.RedisParameterGroupName}}"
      ClusterName: "{{Arguments.Environment}}-redis-{{Arguments.ClusterName}}"
      EngineVersion: "{{Arguments.RedisVersion}}"
      SecurityGroups:
        - Fn::GetAtt: ["RedisSecurityGroup", GroupId]
```

I will provide a PR for this ticket shortly.

PR #494, opened 2017-11-28
#493: Support ClusterName for RedisNode component

inikolaev:
This PR adds support for `ClusterName` property which can be used inside `RedisNode` component definition in order to specify the name of the Redis cluster. If `ClusterName` is not provided the name of the definition will be used instead.

### DIFF
--- a/senza/components/redis_node.py
+++ b/senza/components/redis_node.py
@@ -4,7 +4,7 @@ from senza.utils import ensure_keys
 
 
 def component_redis_node(definition, configuration, args, info, force, account_info):
-    name = configuration["Name"]
+    name = configuration.get("ClusterName", configuration["Name"])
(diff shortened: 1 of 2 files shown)


inikolaev:
@hjacobs Do you have time to take a look? It's a one line change.

hjacobs:
I wonder whether the Senza Redis component still makes sense? It's rather easy to do it with pure Cloud Formation (`Resources` top level key in Senza file).

hjacobs:
:+1:

inikolaev:
@hjacobs The build for Python 3.5 fails with: `ERROR: InvocationError: '/home/travis/build/zalando-stups/senza/.tox/flake8/bin/python setup.py flake8'`
Wonder if it's important or it's a known problem?

inikolaev:
👍

PR #495, opened 2017-12-11
fix senza not to put NAT subnets into server subnets

szuecs:
NAT gateways created by https://github.com/szuecs/kube-static-egress-controller/ will also get a new subnet, because servers could be placed in dmz or private subnets and both has to be routed for given static egress targets.

Senza will normally find these NAT subnets and place ELB interfaces into them, which create failures at runtime.


jmcs:
👍

szuecs:
:+1:

PR #497, opened 2018-01-16
Improve traffic handling

### LINKED ISSUE #490: Stacks may end up with weight of 0 in Route 53
Closely related to #449.

We somehow got a stack with a weight of 0. This is no problem if this is the only stack but when deploying a new stack with `senza create` it will also get a weight of 0 which will cause both stacks to immediately get 50% traffic each which might not be desired.

I could understand this happening when there are two stacks with a 100/0% distribution and the stack with 100% is forcefully deleted but we definitely did NOT do this. We only delete stacks that don't receive any traffic.

As of yet I have no idea how we ended up in this situation. It is extremely undesirable.

jmcs:
Fixes #490

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -20,7 +20,7 @@ import requests
 import yaml
 from botocore.exceptions import ClientError
 from clickclick import (Action, FloatRange, OutputFormat, choice, error,
-                        fatal_error, info, ok)
+                        warning, fatal_error, info, ok)
 from clickclick.console import print_table
 
 from .arguments import (GLOBAL_OPTIONS, json_output_option, output_option,
@@ -46,8 +46,9 @@ from .stups.piu import Piu
 from .subcommands.config import cmd_config
 from .subcommands.root import cli
 from .templates import get_template_description, get_templates
-from .traffic import (change_version_traffic, get_records,
-                      print_version_traffic, resolve_to_ip_addresses)
+from .traffic import (change_version_traffic, get_records, get_weights_for_dns,
+                      print_version_traffic, resolve_to_ip_addresses,
+                      get_stack_versions)
 from .utils import (camel_case_to_underscore, ensure_keys, named_value,
                     pystache_render, get_load_balancer_name)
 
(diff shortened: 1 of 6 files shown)


senza/cli.py
```diff
@@ -592,10 +593,13 @@ def health(region, stack_ref, all, output, field, w, watch):
               help='Ignore failing validation checks')
 @click.option('--update-if-exists', is_flag=True,
               help='Updates the the stack if it exists')
+@click.option('--ensure-no-traffic', is_flag=True,
+              help="Don't create a stack if it would automatically receive "
```
erthalion: I think this phrase fails to deliver the idea that it's happening because of 0 traffic weight - maybe something like "Don't create a stack if it has no traffic and would automatically receive it after the change". What do you think?

senza/traffic.py
```diff
@@ -26,14 +26,10 @@
 DNS_ZONE_CACHE = {}
 
 
-def get_weights(dns_names: list, identifier: str, all_identifiers) -> ({str: int}, int, int):
+def get_weights_for_dns(dns_names):
```
erthalion: Add type hint for `dns_names` ?

senza/cli.py
jmcs: New stacks never have traffic so "if it has no traffic" is always true.

erthalion:
:+1:

senza/cli.py
a1exsh: Is this going to fail when you create the very first version of a stack?

hjacobs: @a1exsh good question..

jmcs: Yes, but the flag is off by default, so you can just omit the flag the first time. I made the behaviour like this to prevent any unexcepted results, specially for Continuous Delivery pipelines.

a1exsh: @jmcs I'm not sure if this buys us a lot if the flag is not enabled by default.  It's easy to forget and if you don't already know about the issue, you have no chance.

Can we do this w/o a flag?  Like always avoid creating a new stack version if there are some existing versions and all of them have "0%" traffic assigned?

hjacobs: @a1exsh do you want to collect some other users upvoting your idea? It would definitely break compatibility, but maybe that's what most Senza users want, so I would be fine with it.

a1exsh: @hjacobs @jmcs Well, I was reluctant to suggest this in the first place, but maybe the better response is to make sure `senza delete` doesn't mess up the traffic weights?

Then you can keep semantics of `senza create` as they are currently and no incompatible change is needed.

Of course one may intentionally set the traffic of the only stack version to 0 and then deploy a second stack, but there is no much sense trying to prevent that.  This will then work as documented by AWS: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-weighted.html#rrsets-values-weighted-weight
> If you set Weight to 0 for all of the records in the group, traffic is routed to all resources with equal probability. This ensures that you don't accidentally disable routing for a group of weighted records.

duergner: I like the "safe by default" approach way better as well. Having an optional flag will require quite some knowledge.

From my experience so far, use cases where the current behavior is desired of the new behavior are rather few actually. Therefore I would be really in favor of a solution which optimises for the 80% even though it would break compatibility. Especially as the broken functionality should be quite obvious (compared to the "broken" behaviour)

a1exsh:
@jmcs is this a "won't fix" now? :)

PR #498, opened 2018-01-26
Add support for multiple channels

aermakov-zalando:
`stups_auto_configuration`: fetch and configure `LatestTaupageStagingImage` and `LatestTaupageDevImage` as well.

### DIFF
--- a/senza/components/taupage_auto_scaling_group.py
+++ b/senza/components/taupage_auto_scaling_group.py
@@ -18,11 +18,6 @@ from senza.utils import ensure_keys
 from typing import Optional
 
 _AWS_FN_RE = re.compile(r"('[{]{2} (.*?) [}]{2}')", re.DOTALL)
-_CHANNEL_MAPPING = {
-    "stable": "LatestTaupageImage",
-    "staging": "LatestTaupageStagingImage",
-    "dev": "LatestTaupageDevImage"
-}
 
 # from kio OpenAPI yaml
 APPLICATION_ID_RE = re.compile(r"^[a-z][a-z0-9-]*[a-z0-9]$")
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Should this maybe be documented in the README?

aermakov-zalando:
@mikkeloscar I'll create a separate PR to [STUPS documentation](https://github.com/zalando-stups/documentation) later. The existing README doesn't have a lot of Taupage documentation, so I don't think it makes sense to document this change here.

mikkeloscar:
:+1:

senza/components/taupage_auto_scaling_group.py
```diff
@@ -18,6 +18,11 @@
 from typing import Optional
 
 _AWS_FN_RE = re.compile(r"('[{]{2} (.*?) [}]{2}')", re.DOTALL)
+_CHANNEL_MAPPING = {
+    "stable": "LatestTaupageImage",
```
jmcs: Are the names `LatestTaupageImage`, `LatestTaupageStagingImage` and `LatestTaupageDevImage` used outside senza code? Why not just use `stable`, `staging` and `dev`?

aermakov-zalando: `LatestTaupageImage` is used written into the generated CF template, so it's possible that someone else depends on it. I don't think it's a good idea to break backwards compatibility over a trivial rename.

szuecs:
@jmcs does it requires another change?
For me this looks good

aermakov-zalando:
I've removed the `TaupageChannel` property from the `TaupageAutoScalingGroup` component. `Image: LatestTaupageStagingImage` is a bit more verbose, but this is an already documented way of selecting an AMI.

aermakov-zalando:
👍

jmcs:
👍

PR #501, opened 2018-02-15
Add hyphen support for versions besides alphanumerical characters

evangelion1204:
Hi,

We would need support for also hyphens in the version and there seems to be no limitation from AWS.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -184,7 +184,7 @@ class KeyValParamType(click.ParamType):
 
 def validate_version(ctx, param, value):
     if not VERSION_PATTERN.match(value):
-        raise click.BadParameter('Version must satisfy regular expression pattern "[a-zA-Z0-9]+"')
+        raise click.BadParameter('Version must satisfy regular expression pattern "[a-zA-Z0-9-]+"')
     return value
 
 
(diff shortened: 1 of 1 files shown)


mikkeloscar:
I think senza (and maybe other components) treats the last part of the stackname as version e.g. `name-<version>` So I suspect having dashes in the version will cause problems where this convention is expected.

evangelion1204:
That might be true, but versions need to match the docker tags that are pulled for deployment. I would like to do something like feature-branch based deployments on PRs. So I need to add either the PR id or branch-name either to the version (seems to be clean) or to artefact name which would spam our artefacts.

hjacobs:
👎

evangelion1204:
Maybe what I want to achieve is a kind of hack. But as long as stackname is bound directly to the artifact this will always ne an issue or is there any workaround?

evangelion1204:
Solved via senza definition `source` taupage config.

hjacobs:
@evangelion1204 I think there is a misunderstanding: your stack name does not need to be bound to the artifact version! You can use arbitrary stack names (like "v1", "v2" or even "green", "blue") and have semantic Docker artifact versions like "1.2.0-13".

PR #503, opened 2018-03-13
Add support for release channels in senza images/senza patch

aermakov-zalando:
* `senza image` now shows AMIs in all 3 channels, even if they're not used
 * `senza patch` now supports `--image staging` and `--image dev` in addition to `--image latest`


jmcs:
:+1:

PR #509, opened 2018-05-15
Cleanup of pierone CVE analysis

fatroom:
Due to drop of pierone clair integration there's no more sense to keep the warnings during senza deployment (since pierone will reply with NOT_PROCESSED_YET for every image).

Basically this reverts #226


jmcs:
:+1:

dryewo:
:+1:

PR #514, opened 2018-07-04
Extend spotinst elastigroup

lmineiro:
This change tries to simplify the definition of Elastgroups by assuming some defaults like the native Auto Scaling Group counterpart.

It also adds support for the automatic subnet mapping from `Senza::StupsAutoConfiguration`

It assumes, for now, that the cloud accounts provisioned in Spotinst will be named "aws:<aws-account-id>". This is used for Spotinst cloud account resolution and it's needed to use a specific account instead of the default one.

/cc @danieldop @jmcs

### DIFF
--- a/spotinst/components/elastigroup.py
+++ b/spotinst/components/elastigroup.py
@@ -131,7 +131,7 @@ def extract_subnets(definition, elastigroup_config, account_info):
     """
     subnet_ids = elastigroup_config["compute"].get("subnetIds", [])
     target_region = elastigroup_config.get("region", account_info.Region)
-    if len(subnet_ids) < 1:
+    if not subnet_ids:
         subnet_ids = [subnetId for subnetId in
                       definition["Mappings"]["ServerSubnets"].get(target_region, {}).get("Subnets", [])]
     elastigroup_config["region"] = target_region
--- a/spotinst/components/elastigroup.py
+++ b/spotinst/components/elastigroup.py
@@ -53,7 +53,7 @@ def component_elastigroup(definition, configuration, args, info, force, account_
         "Properties": {
             "ServiceToken": create_service_token(args.region),
             "accessToken": access_token,
-            "accountId": extract_spotinst_account_id(access_token, definition),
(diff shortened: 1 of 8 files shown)


spotinst/components/elastigroup.py
```diff
@@ -17,23 +34,148 @@ def component_elastigroup(definition, configuration, args, info, force, account_
...
+
+
+def extract_subnets(definition, elastigroup_config, account_info):
+    """
+    This fills in the subnetIds and region attributes of the Spotinst elastigroup, in case their not defined already
+    The subnetIds are discovered by Senza::StupsAutoConfiguration and the region is provided by the AccountInfo object
+    """
+    subnet_ids = elastigroup_config["compute"].get("subnetIds", [])
+    target_region = elastigroup_config.get("region", account_info.Region)
+    if len(subnet_ids) < 1:
```
jmcs: `if not subnet_ids` would have the same effect while avoid the `len` call

jmcs:
:+1:

lmineiro:
:+1:

lmineiro:
Thanks for the review @talzur.

> spotinst/components/elastigroup.py:112 - why are the tags only added when none exist?
This is for consistency. Across pretty much every Elastigroup settings, whenever they're defined already, those are taken as the desired state and left untouched. For the particular case of the tags, the expectation is that if someone defined custom tags, any of the 3 "automatic" tags can also be defined there only once, which seems like a reasonable effort.

> spotinst/components/elastigroup.py:123 - is this the only place you set the group name?
Great catch. The Elastigroup name should always be defined, regardless of having tags defined or not. I created https://github.com/zalando-stups/senza/issues/518 to track this.

> spotinst/components/elastigroup.py:21 - ELASTIGROUP_DEFAULT_PRODUCT
Thanks for the thorough explanation. To the best of my knowledge, the entire [STUPS platform](https://stups.io/), which Senza is part of, was created for VPC setups. This should be fine. I know it will be for our usage, at least.

PR #515, opened 2018-07-05
Senza kms decryption

### LINKED ISSUE #513: KMS decryption before CloudFormation Orchestration
For senza files with sensitive parameters, those parameters would ideally be encrypted.

It is possible to use KMS encryption to avoid having those values as plain text in the senza definition file. This, however, means that those parameters must be decrypted before CloudFormation orchestration, and it will happen on a local machine running senza, or through lizzy.

The suggestion is to use the prefix `senza:kms:` to clearly identify these parameters.
For reasons of simplicity, those parameters should be in the `SenzaInfo` section.

Ex.:

```
SenzaInfo:
  SensitiveParameter: "senza:kms:skdfjg-sdFDGS23rwfkgjhsdfGSDGER034teFSD"

SenzaComponents:
(...)
      someParameter: {{SenzaInfo.SensitiveParameter}}
```

A concrete example for this functionality is ElastiGroup's `SpotinstAccessToken` parameter: https://github.com/zalando-stups/senza/blob/master/examples/elastigroup.yaml#L5

The inspiration for this approach comes from Taupage's use of KMS encryption. In that case the prefix is `aws:kms:`

pc-alves:
Closes #513

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -219,7 +219,14 @@ BASE_TEMPLATE = {
 
 
 def decrypt_parameters(definition):
-    senza_info = definition["Mappings"]["Senza"]["Info"]
+    """
+    Searches for encrypted values in the SenzaInfo properties and decrypts them.
+    """
+    try:
+        senza_info = definition["Mappings"]["Senza"]["Info"]
+    except KeyError:
+        return definition
+
     for key in senza_info:
         if str(senza_info[key]).startswith(SENZA_KMS_PREFIX):
             senza_info[key] = decrypt_kms(senza_info[key])
@@ -228,6 +235,9 @@ def decrypt_parameters(definition):
 
 
 def decrypt_kms(val):
+    """
+    Decrypts KMS encrypted values. KMS encrypted values have the prefix 'senza:kms:'
+    """
     ciphertext_blob = val[len(SENZA_KMS_PREFIX):]
     ciphertext_blob = base64.b64decode(ciphertext_blob)
     kms_client = boto3.client(service_name='kms', region_name=get_region(region=None))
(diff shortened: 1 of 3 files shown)


senza/cli.py
```diff
@@ -217,6 +218,24 @@ def watching(w: bool, watch: int):
 }
 
 
+def decrypt_parameters(definition):
```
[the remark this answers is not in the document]
jmcs: Can you had a docstring, explaining what the function does, noting that only values in Senza Info are decrypted?

senza/cli.py
```diff
@@ -217,6 +218,24 @@ def watching(w: bool, watch: int):
...
+    senza_info = definition["Mappings"]["Senza"]["Info"]
```
jmcs: This should be something like:

```python
try:
    senza_info = definition["Mappings"]["Senza"]["Info"]
except KeyError:
    return definition
```
to avoid crashing if SenzaInfo is missing.

senza/error_handling.py
```diff
@@ -103,6 +103,11 @@ def die_unknown_error(self, unknown_exception: Exception):
                             'content of {fn}'.format(e=unknown_exception,
                                                      fn=file_name))
 
+    # TODO :: sounds like I will have to do something here
+    # InvalidCiphertextException
+    # KeyUnavailableException
```
jmcs: I think those are going to be `ClientError` and we can skip them for now, since the only consequence is an ugly error message.

pc-alves: Fair enough. I'll add the exception handling as a TODO in the issue.

jmcs:
The tests are failing. You need to pass a region (even a fake one) to the tested commands.

jmcs:
:+1:

pc-alves:
👍

PR #516, opened 2018-07-10
added missing functionality to elastigroup

lmineiro:
This should bring Spotinst's Elastigroup to feature parity with native AWS Auto Scaling Groups.

It also changes the syntax to reuse all of the ASG settings while allowing the more advanced Elastigroup API definition, which takes precedence whenever defined.

### DIFF
--- a/examples/elastigroup.yaml
+++ b/examples/elastigroup.yaml
@@ -9,18 +9,14 @@ SenzaComponents:
 
   - AppServerElastigroup:
       Type: Spotinst::Elastigroup
-      Elastigroup:
-        compute:
-          instanceTypes:
-            ondemand: m3.large
-            spot:
-              - m3.large
-              - m4.large
-              - c3.large
-              - c4.large
-          launchSpecification:
-            SecurityGroups: default # for this to work make sure the default security group allows ingress on TCP/8000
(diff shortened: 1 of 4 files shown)


lmineiro:
:+1:

lmineiro:
Thanks again @talzur 

> spotinst/components/elastigroup.py:361 - Perhaps you would want to use two kinds of defaults here

This was an oversight. The API clearly shows those different healthcheck types.

> spotinst/components/elastigroup.py:444-448 If I am not mistaken the code here can lead to a situation where you do not set up an OD type. Created #519 for tracking.

This was yet another oversight. We thought the On Demand was actually optional and programmed it according to that assumption. Thanks for confirming that's not the case. Created #520 for tracking.

aharont:
Hi @lmineiro , not sure I understand the TargetGroup handling:
spotinst/components/elastigroup.py:362
When defining the target groups entities in the template should we name them with 'TargetGroup' suffix or without it?
Then how should we use them in the Elastiroup template? with the suffix or without it? 
I'm afraid that if we'll refer it with the suffix and then add another 'TargetGroup' suffix like the code in L362 does the CFN template won't find the target group entity.
If we won't refer it with the suffix then the Senza template won't find the target group entity.

Issue #517, opened 2018-07-11
Add support Elastigroup for all commands


lmineiro:
Some command line options don't work with Spotinst Elastigroups, only native Auto Scaling Groups.

The commands that seem to work:
create
list
domains
dump
events
health
images
resources
traffic
update

The commands that don't work or don't offer Elastigroup:
init - no knowledge about Elastigroups
instances - no error, just no instances listed
patch - no error and no action
respawn-instances - no error and no action
scale - no error and no action
status - number of reported instances is wrong

pc-alves:
Scale updated to support Elastigroup

dmytro-vynokurov:
@lmineiro @pc-alves are there any plans on adding support for the `instances` command to work with ElastiGroup?

lmineiro:
According to my own experiments, the time the command takes, even without `--all`, varies a lot and ca be very slow at time.

My test using the stack reference:
```
$ time senza instances skipper
Stack Name│Ver.│Resource ID│Instance ID        │Public IP     │Private IP   │State  │LB Status│Launched
<28 instances listed>
real	0m23.177s
user	0m2.429s
sys	0m0.265s
```

Followed almost immediately by the same query with `--all`
```
$ time senza instances skipper --all
Stack Name│Ver.│Resource ID│Instance ID        │Public IP     │Private IP   │State  │LB Status│Launched
<the same 28 instances listed>
real	0m15.200s
user	0m2.342s
sys	0m0.244s
```

It's just a workaround until we're confident that something like https://github.com/zalando-stups/senza/pull/547 is an acceptable solution.

dmytro-vynokurov:
@lmineiro thank you for the explanation. Then I guess we'll stick to the workaround until senza gets updated

dmytro-vynokurov:
@lmineiro done. It works. Thanks a lot for implementing this :smile:

PR #525, opened 2018-07-19
Scale for elastigroup

pc-alves:
Contributes to #517


senza/cli.py
```diff
@@ -25,7 +25,7 @@
                         fatal_error, info, ok)
 from clickclick.console import print_table
 
-import spotinst
+from spotinst.components import elastigroup_api
```
jmcs: This should be `from .spotinst.components import elastigroup_api` since it's a relative import.

jmcs:
:+1:

senza/cli.py
pc-alves: This won't work because the spotinst package is outside the senza package. Let's merge this one as a absolute import, and move the spotinst package inside senza in the next PR (there are a few more functionalities to be changed)

pc-alves:
👍

Issue #526, opened 2018-07-20
Support traffic switching for stacks with multiple load balancers


adetunjiakintundeakinde:
The variable partial_sum and partial count is miscalculated  for stacks with multiple load balancers

PR #527, opened 2018-07-20
#526 fixed calculation of partial count and sum

adetunjiakintundeakinde:
Traffic switching was not working for stacks with multiple load balancers.The partial sum and count variables were incorect


jmcs:
Can you solve the issues pointed out by codacy? It should be enough to make the build pass.

jmcs:
:+1:

adetunjiakintundeakinde:
👍

PR #528, opened 2018-07-24
refactored Spotinst cloud account matching

lmineiro:
Spotinst changed the API to support backend [matching for a given AWS account ID](https://api.spotinst.com/organization-accounts/get-accounts/)

This change makes the cloud account resolution more efficient and less fragile (no longer parsing the cloud account name).

This should also enable users to keep more user-friendly names in the Spotinst cloud accounts.


lmineiro:
:+1:

jmcs:
:+1:

PR #530, opened 2018-07-31
Respawn and patch for elastigroup

pc-alves:
Contributes to #517 

Adds patch and respawn functionality to elastigroup stacks


jmcs:
:+1:

pc-alves:
👍

PR #535, opened 2018-09-04
removed Config suffix from Elastigroup resource names

lmineiro:
This shouldn't cause any surprises since the previous behavior was "mostly" to keep the explicit resource names.

It closes #534 

/cc @pc-alves @jmcs

PR #538, opened 2018-09-27
Fix spotinst stack details

lmineiro:
This should fix the problem to use the auxiliary commands scale, patch and respawn
It stopped working with #535 and would fail if the Elastigroup component name was not `AppServer`

### DIFF
--- a/senza/spotinst/components/elastigroup_api.py
+++ b/senza/spotinst/components/elastigroup_api.py
@@ -6,7 +6,7 @@ import requests
 import json
 import boto3
 
-from components.elastigroup import ELASTIGROUP_RESOURCE_TYPE
+from senza.components.elastigroup import ELASTIGROUP_RESOURCE_TYPE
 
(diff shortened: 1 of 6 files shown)


lmineiro:
:+1:

senza/spotinst/components/elastigroup_api.py
```diff
@@ -22,16 +24,22 @@ def __init__(self, account_id, access_token):
 
 
 def get_spotinst_account_data(region, stack_name):
-    '''
-    Extracts required parameters required to access SpotInst API
-    '''
+    """
+    Extracts the Spotinst API access token and cloud accoint ID required to use the SpotInst API
+    It returns those parameters from the first resource of Type Customer::elastigroup
```
pc-alves: Two typos:
 - 'cloud account ID'
 - 'resource of type Custom::elastigroup'

pc-alves:
👍

lmineiro:
:+1:

PR #539, opened 2018-09-28
trying to fix flaky test_invalid_definition

lmineiro:
In Travis this test contains an upper case `DEFINITION`. CDP builds have a lower case `definition`.
The case should not be relevant to the semantics of the test.


lmineiro:
:+1:

jmcs:
👍

Issue #540, opened 2018-10-02
Wrong base64 encoding for userdata


lmineiro:
The `Senza::Elastigroup` component encodes the EC2 userdata script using `base64.urlsafe_b64encode()`. When the input contains the character `>`, the creation of the Elastigroup fails.

This happens because the URL implementation of [Base64 encoding](https://tools.ietf.org/html/rfc4648#section-5) replaces the 62nd character with the character `-` instead of the expected `+`.

Botocore uses a [different approach](https://github.com/boto/botocore/blob/cbd5b272205fb85532f067c9fe40a0b1bb21be25/botocore/handlers.py#L531). This is the true base64 encoding.

Note: The patch command also uses the `base64.urlsafe_b64encode()` function.
A brief test didn't show any error when using AWS Auto Scaling Group but it also caused the Elastigroup call to fail.

PR #541, opened 2018-10-04
fixed the user data encoding by using the proper base64

lmineiro:
The function `base64.urlsafe_b64encode()` would fail to properly encode the `>` character, as stated in #540 

That affected the Elastigroup creation whenever the `>` was present in the user data.

For the CloudFormation template the fix was to use the same deferred call to `Fn::Base64`, just [like for the native Auto Scaling Groups](https://github.com/zalando-stups/senza/blob/0b7c7cee8d7c4f1a2bc01c9cc2cad42171292d98/senza/components/taupage_auto_scaling_group.py#L144).

The patch command was changed to use the correct `base64.b64encode()` function.

### DIFF
--- a/senza/components/elastigroup.py
+++ b/senza/components/elastigroup.py
@@ -315,8 +315,8 @@ def extract_user_data(configuration, elastigroup_config, info: dict, force, acco
         if not force and docker_image.registry:
             check_docker_image_exists(docker_image)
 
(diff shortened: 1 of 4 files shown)


pc-alves:
👍

jmcs:
:+1:

PR #546, opened 2018-10-24
Fix flake8 checks

jmcs:
The newest release of flake8 is stricter than previous ones, so the master build was broken.

This PR fixes it by:
- Formatting affected files with black;
- Removing an unused assignment.


lmineiro:
:+1:

PR #547, opened 2018-10-25
change in the strategy to report instances

lmineiro:
This small change seems to be enough to allow reporting instances managed by Elastigroups.

Only AWS can set `aws:` prefixed tags. Instances created by the Elastigroups don't have the `aws:cloudformation:stack-name` tag that was used to filter instances.

Pragmatically, I believe that the `Name` tag contains exactly the same information, although the guarantee that it was created by CloudFormation no longer exists.

If this new approach doesn't sound reasonable, please share the particular uses cases or assumptions where this change will break.

This also changed the instance state filtering to the server side for some micro optimization.

This updates #517

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1271,8 +1271,13 @@ def instances(
     if all:
         filters = []
     else:
-        # filter out instances not part of any stack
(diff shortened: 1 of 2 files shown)


lmineiro:
:+1:

jmcs:
:+1:

PR #553, opened 2018-11-23
Always add standard STUPS EC2 tags

inikolaev:
This PR makes changes to the way standard STUPS tags specified within `ElastiGroup` component are handled:

* Standard tags are always added to the list of tags even in the presence of custom tags
* Any standard tags specified in configuration are overridden by values from Senza definition - I have never ever had to specify these tags manually, so please let me know if you think there's a valid use case when it might be useful to actually override tags and I will change the logic

### DIFF
--- a/senza/components/elastigroup.py
+++ b/senza/components/elastigroup.py
@@ -249,17 +249,29 @@ def fill_standard_tags(definition, elastigroup_config):
     Elastigroup name attribute to the same value as the EC2 Name tag if found empty.
     The default STUPS EC2 Tags are Name, StackName and StackVersion
     """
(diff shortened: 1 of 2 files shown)


lmineiro:
:+1:

lmineiro:
fyi, not setting the tags whenever the property was present was a deliberate decision to stay consistent. The behavior for the remaining properties is also to leave them untouched if present in the definition.
The reasoning for that is that it's possible to set them if wanted but also possible NOT to set them. With your proposal, it will no longer be possible NOT to set them.

I think that's still reasonable as the general behavior of Senza when using native ASGs is to set them

inikolaev:
@lmineiro so, should we set them as per this PR? I'm confused :) 
I could add some boolean property which would be used to disable this, but default would be to always set STUPS tags.

pc-alves:
👍

PR #554, opened 2018-11-29
change in the strategy to report instance status

lmineiro:
This follows the same pragmatic approach as #547 

It updates #517


lmineiro:
:+1:

pc-alves:
👍

PR #556, opened 2018-12-18
docker image tags tests fixes

olevchyk:
see https://github.com/zalando-stups/pierone-cli/pull/78 
there is no `get_image_tag` in api.py after above PR. 
Switched to get_latest_tag that does almost the same.


jmcs:
:+1:

pc-alves:
👍

PR #557, opened 2019-01-10
LB health checks configurations. 

olevchyk:
This PR adds health checks related settings to be configurable from `AppLoadBalancer` senza section. 
Found that we can set `HealthCheckGracePeriod` for AutoScalingGroup to avoid failing of long-starting apps, 
but still having these settings configurable would something nice to have.


olevchyk:
failing tests fixed in https://github.com/zalando-stups/senza/pull/556

PR #559, opened 2019-03-21
showing details about failed respawn/deploy in cli

### LINKED ISSUE #558: senza respawn and deployment with Elastigroup blocked by previously failed deployment
When in the past a deployment of an Elastigroup failed, the deployment does not reach a final state, it is still running until stopped manually. Due to this a new deploy or respawn will fail with "Can't have 2 Rolls at the same time. Please stop the previous one."

These kind of errors are not handled in or forwarded to the cli, the user just gets "Unknown Error: 400 Client Error" and is clueless about what happened.

cjander:
closes #558


hjacobs:
This project is not actively maintained anymore (#573). Closing.

PR #562, opened 2019-05-08
Add Cross Stack Policy to IAM Role

pc-alves:
On March 18, 2019, AWS introduced limitations to the scope of calls to stack level APIs, such as DescribeStackResources. These APIs are used by aws-cfn-bootstrap scripts, including cfn-init and cfn-hup, which are commonly used to provision EC2 instances created by CloudFormation.

This limitation has not been applied to accounts that are making cross stack requests, but this will only be like that until May 30, 2019. When this comes into effect, it will prevent Elastigroup deployments from functioning as expected.

The policy added in this PR is the one recommended by Spotinst [in their documentation](https://api.spotinst.com/provisioning-ci-cd-sdk/provisioning-tools/cloudformation/cfn-helper/).

This policy will be added when running the `init` command and when creating new Elastigroup stacks via `create` command. The policy will be added only if not found in the provided specification.

Current TODOs:

- [x] Fix issue with imports in tests
- [x] Patch policy in `create` command

### DIFF
--- a/senza/templates/_helper.py
+++ b/senza/templates/_helper.py
@@ -3,9 +3,12 @@ import re
 
 import boto3
 import botocore.exceptions
+# TODO :: WITHOUT THIS IMPORT, TEST test_template_helper_check_iam_role CANNOT MOCK click.confirm
 import click
 import clickclick
 from clickclick import Action
+# TODO :: WITHOUT THIS IMPORT, TESTS test_init, test_init_opt2 and test_init_opt5 FAIL DUE TO IMPORT ERROR
+from click import confirm
 from senza.aws import get_account_alias, get_account_id, get_security_group
 
 from ..manaus.boto_proxy import BotoClientProxy
--- a/senza/templates/_helper.py
+++ b/senza/templates/_helper.py
@@ -167,6 +167,8 @@ def check_iam_role(application_id: str, bucket_name: str, region: str):
         ],
         "Version": "2008-10-17",
     }
+
+    create = False
     if not exists:
         create = confirm(
             "IAM role {} does not exist. "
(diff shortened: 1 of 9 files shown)


senza/templates/_helper.py
```diff
@@ -180,20 +212,36 @@ def check_iam_role(application_id: str, bucket_name: str, region: str):
...
+            mint_read_policy = create_mint_read_policy_document(application_id, bucket_name, region)
             iam.put_role_policy(
                 RoleName=role_name,
                 PolicyName=role_name,
-                PolicyDocument=json.dumps(policy),
+                PolicyDocument=json.dumps(mint_read_policy),
+            )
+
+    cross_stack_policy_exists = False
+    cross_stack_policy_name = "CrossStackPolicy-{}".format(application_id)
```
lmineiro: Let's use the name `system-cf-notifications` for the policy to match other existing tools

tests/test_manaus/test_acm.py
```diff
@@ -218,7 +218,9 @@ def test_get_certificates(monkeypatch):
     assert len(certificates_net) == 1
     assert certificates_net[0].arn == 'arn:aws:acm:eu-west-1:cert2'
 
-    m_client.describe_certificate.side_effect = [{'Certificate': CERT3}]
+    # TODO :: NOT MEANT TO BE MERGED!!! JUST HERE TO FIX THE TEST IN PYTHON 3.7
```
lmineiro: Do not forget this ;-)

senza/templates/_helper.py
pc-alves: Changed

tests/test_manaus/test_acm.py
pc-alves: Turns out that this was actually the best way to fix the test, so I removed the comment

senza/templates/_helper.py
```diff
@@ -180,20 +211,47 @@ def check_iam_role(application_id: str, bucket_name: str, region: str):
...
+                "IAM role {} already exists. ".format(role_name)
+                + "Do you want Senza to overwrite the role policy?"
+            )
         )
     )
-    if update_policy:
-        with Action("Updating IAM role policy of {}..".format(role_name)):
-            policy = get_iam_role_policy(application_id, bucket_name, region)
+    if attach_mint_read_policy:
+        with Action("Updat"
```
jmcs: Please remove `/n`

pc-alves: Done

lmineiro:
👍

pc-alves:
👍

PR #567, opened 2019-05-24
no longer poping the SpotinstAccessToken attribute

### LINKED ISSUE #566: Allow multiple Elastigroup components
When deploying `Senza::Elastigroup` components the `SpotinstAccessToken` property [is being removed](https://github.com/zalando-stups/senza/blob/e9f84724628b4761f8d5da4d37a2993f11d6433b/senza/components/elastigroup.py#L595) from the SenzaInfo dictionary.

If there are multiple `Senza::Elastigroup` components, only the first one will succeed and the remaining fail due to the missing attribute.

lmineiro:
This fixes #566

### DIFF
--- a/senza/components/elastigroup.py
+++ b/senza/components/elastigroup.py
@@ -592,7 +592,7 @@ def _extract_spotinst_access_token(definition: dict):
     """
     extract the provided access token
     """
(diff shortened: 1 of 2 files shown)


lmineiro:
:+1:

pc-alves:
👍

PR #569, opened 2019-05-27
Rolling back changes introduced for the cross stack policy

pc-alves:
After additional tests and experiments it was determined that Elastigroups do not really need the cross stack policy.
The changes to senza's `create` and `init` commands, introduced with #562 are then rolled back.


lmineiro:
👍

pc-alves:
👍

PR #571, opened 2020-01-24
overwrite minSize instances if 0

mschwitalla:
If the `MinSize` number of the autoscaling group was 0 and the user wants to scale up, we are overwriting `MinSize` with the new desired capacity. Users might not be aware of the fact that their stack might be downscaled after a time.

### DIFF
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1964,7 +1964,7 @@ def respawn_instances(stack_ref, inplace, force, batch_size_percentage, region):
 @click.option(
     "--min_size",
     metavar="MIN_SIZE",
-    type=click.IntRange(0,100, clamp=True),
+    type=click.IntRange(0, 100, clamp=True),
     help="Define the MinSize to update the AutoScalingGroup.")
 @click.option(
     "-f", "--force", is_flag=True, help="Force scaling multiple stacks if needed"
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -2047,6 +2047,12 @@ def scale_auto_scaling_group(asg, asg_name, desired_capacity):
             act.ok("NO CHANGES")
         else:
             kwargs = {}
+            if group["MinSize"] == 0 and desired_capacity > group["MinSize"]:
+                kwargs["MinSize"] = desired_capacity
+                info(
+                    "Min instance number was set to 0. Overwriting it with "
+                    "desired capicity of {} to prevent unexpected downscaling.".format(desired_capacity)
+                )
             if desired_capacity < group["MinSize"]:
                 kwargs["MinSize"] = desired_capacity
             if desired_capacity > group["MaxSize"]:
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -1962,7 +1962,7 @@ def respawn_instances(stack_ref, inplace, force, batch_size_percentage, region):
 @click.argument("stack_ref", nargs=-1)
 @click.argument("desired_capacity", type=click.IntRange(0, 100, clamp=True))
 @click.option(
-    "--min_size",
+    "--min-size",
     metavar="MIN_SIZE",
     type=click.IntRange(0, 100, clamp=True),
     help="Define the MinSize to update the AutoScalingGroup.")
--- a/senza/cli.py
+++ b/senza/cli.py
@@ -2051,7 +2051,7 @@ def scale_auto_scaling_group(asg, asg_name, desired_capacity):
                 kwargs["MinSize"] = desired_capacity
                 info(
                     "Min instance number was set to 0. Overwriting it with "
-                    "desired capicity of {} to prevent unexpected downscaling.".format(desired_capacity)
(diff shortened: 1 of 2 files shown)


senza/cli.py
```diff
@@ -2047,6 +2047,12 @@ def scale_auto_scaling_group(asg, asg_name, desired_capacity):
             act.ok("NO CHANGES")
         else:
             kwargs = {}
+            if group["MinSize"] == 0 and desired_capacity > group["MinSize"]:
+                kwargs["MinSize"] = desired_capacity
+                info(
+                    "Min instance number was set to 0. Overwriting it with "
+                    "desired capicity of {} to prevent unexpected downscaling.".format(desired_capacity)
```
lmineiro:
```suggestion
                    "desired capacity of {} to prevent unexpected downscaling.".format(desired_capacity)
```

senza/cli.py
```diff
@@ -2047,6 +2047,12 @@ def scale_auto_scaling_group(asg, asg_name, desired_capacity):
             act.ok("NO CHANGES")
         else:
             kwargs = {}
+            if group["MinSize"] == 0 and desired_capacity > group["MinSize"]:
```
lmineiro: This is a considerable change in the behavior. It also may result in some nasty side-effects for operations such as `senza scale 100`. Have you considered just logging the warning? That would already help operators to realize that `MinSize` is left at 0. A new flag, such as `--force` could then be used to do what you suggest. This would keep the backwards compatibility for any existing scripts but also give a hint to operators how to achieve the new desired behavior.

mschwitalla: Is there a way to update/patch the `MinSize` value of the ASG through senza? If yes, then we could add it to the message.

lmineiro: What I was suggesting would be some sort of new flag for the scale command.
e.g. `senza scale --force foo 10`
The warning message would then suggest this flag whenever the MinSize was found at 0. That way the operator would know exactly what to do if that was intended. With time, the flag would become known along with its effects and new scripts that want the new behavior could just add the flag.

lmineiro: Just realized that `--force` already exists for scale. It'd have to be another flag

mschwitalla: The scale command already contains a `--force` flag. It was up until now only used for updating more stacks than 1.

I would propose to use the flag also for this use case

lmineiro: What do you think about `--minsize [value]`? It could work with or without arguments. With an argument it'd allow the operator to specify a minimum capacity different from the max size. Without the argument it'd use the same value as max size.

mschwitalla: I added the option `--min_size` that requires a value. I think this makes it the most explicit for users. 

What do you think?

senza/cli.py
```diff
@@ -1961,12 +1961,17 @@ def respawn_instances(stack_ref, inplace, force, batch_size_percentage, region):
 @cli.command()
 @click.argument("stack_ref", nargs=-1)
 @click.argument("desired_capacity", type=click.IntRange(0, 100, clamp=True))
+@click.option(
+    "--min_size",
+    metavar="MIN_SIZE",
+    type=click.IntRange(0,100, clamp=True),
```
jmcs: Codacy Issue found: [Exactly one space required after comma](https://app.codacy.com/manual/zalando-bus/senza/pullRequest?prid=4882207)

jmcs:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/zalando-bus/senza/pullRequest?prid=4882207&bid=16176928)

mschwitalla:
@lmineiro @jmcs The build is failing for some reason. Looking at travis it seems to be related to some errors in the setup? Can you help? 

Running the tests locally works

senza/cli.py
```diff
@@ -1961,12 +1961,17 @@ def respawn_instances(stack_ref, inplace, force, batch_size_percentage, region):
 @cli.command()
 @click.argument("stack_ref", nargs=-1)
 @click.argument("desired_capacity", type=click.IntRange(0, 100, clamp=True))
+@click.option(
+    "--min_size",
```
hjacobs: please use `--min-size` for consistency

tests/test_cli.py
```diff
@@ -1788,6 +1788,53 @@ def test_scale_with_force_confirm(monkeypatch):
                            catch_exceptions=False)
     assert 'Scaling myasg from 1 to 2 instances' in result.output
 
+def test_scale_with_overwriting_zero_minsize(monkeypatch):
+  boto3 = MagicMock()
```
hjacobs: please use 4-spaces as indent (PEP8)

mschwitalla:
Is anybody able to help me with the build pipeline?

jmcs:
@mschwitalla can you rebase your branch on the latest master?

mschwitalla:
@jmcs done

jmcs:
:+1:

mschwitalla:
👍

PR #579, opened 2023-02-21
Always set HttpTokens to optional

hughcapet:
This commit forces MetadataOptions->HttpTokens to always be set to 'optional' by senza, as we want to still be able to use IMDSv1, while base AMIs enforce v2 and ebs packer still doesn't allow this to be redefined.


Jan-M:
:+1:

hughcapet:
👍

PR #580, opened 2023-02-28
Fix testing pipeline

hughcapet:
- CDP overlays -> runtime
- Remove top-level region option
- Define a separate Command class for flake8 in setup.py (flake8 setuptools/setup.py support is now deprecated)
- Fix a flake8 error

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -6,7 +6,9 @@ notifications:
 pipeline:
   - id: build
     type: script
-    overlay: ci/python
+    vm_config:
+      type: linux
+      image: cdp-runtime/python-3.9
     commands:
     - desc: "Install dependencies"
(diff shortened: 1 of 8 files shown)


Jan-M:
:+1:

jmcs:
LGTM but I've a question. Why are you removing the top-level region option?

hughcapet:
@jmcs 

sorry, should have posted the message from one of the commits. This is because this was causing the following misbehavior (revealed in the `test_cli.py` unit tests):
```
The current implementation uses value set via the envvar AWS_DEFAULT_REGION
for sub-commands even if an explicit flag value is provided.
```

And I actually don't fully understand the whole idea of having the top-level region option.

hughcapet:
👍

